PR #14, opened 2025-03-28
fix: ollama chat using stream

andrewbroberg:
Fixes Ollama chat function using stream response due to the array_filter removing the stream: false property

I have allowed Ollama to receive an option Guzzle Client property so that the request/response can be mocked and tested properly.  I've left url as is for now as it was to avoid backwards compatibility breaks, happy to discuss a different way to handle this approach if you want.


ilvalerione:
Hi @andrewbroberg 
Thank you so much for the contribution!

I think a method "setClient" could be added into the class to avoid collecting too much things in the constructor. I would like to keep the constructor clean because in my mind the constructor parameters serve like an intuitive way to make developers aware of the basic parameters needed to create the instance inside the Agent class.

andrewbroberg:
@ilvalerione That sounds like a good idea, I'll make that change now.

andrewbroberg:
@ilvalerione Updated. Have added it as a Trait as I imagine it's going to be reused with other Providers

ilvalerione:
It was exactly my intention. Thanks

ilvalerione:
Released as 1.8.6

https://github.com/inspector-apm/neuron-ai/releases/tag/1.8.6

PR #15, opened 2025-03-28
Pass additional/custom params to OpenAI chat/completions endpoint

iztok:
With this change we can 
- add additional params to chat API
- use $result['choices'][0]['message']['parsed'] in case we get back structured response

### DIFF
--- a/src/Providers/OpenAI/HandleChat.php
+++ b/src/Providers/OpenAI/HandleChat.php
@@ -27,6 +27,7 @@ trait HandleChat
         $json = [
             'model' => $this->model,
             'messages' => $mapper->map(),
+            ... $this->additional_params,
         ];
 
         // Attach tools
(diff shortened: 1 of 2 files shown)


ilvalerione:
I didn't find any mention of the "parsed" field in the message response. Did you find it during live test? Or is there e reference somewhere?

ilvalerione:
I tested your example, but it never returns `$result['choices'][0]['message']['parsed']`

Also calling a tool it works with default messages. Let me know if you have any additional details, otherwise I'm going to close the PR, and continue to work on my implementation.

iztok:
Correct, the `$result['choices'][0]['message']['parsed']` is not part of the response. I removed that section of code.

The `$additional_params` part of this PR does solve multiple issues:
- set structured output (my initial motivation)
- set temperature
- set max_completion_tokens
- etc [https://platform.openai.com/docs/api-reference/chat/create](https://platform.openai.com/docs/api-reference/chat/create)

Not sure if it's on your backlog to let pass additional custom parameters to providers by some other way.

ilvalerione:
I close the PR because in the meantime I added other things for the next release. But basically it will contain this proposal. Thank you.

Issue #19, opened 2025-04-04
setInput fails when all parameters are nullable


marcelfolaron:
**Describe the bug**
I am currently building an agent using Anthropic and have various methods where all parameters can be null. 
If claud decides that no parameters are necessary I get the error. 

`[object] (TypeError(code: 0): NeuronAI\Tools\Tool::setInputs(): Argument #1 ($inputs) must be of type array, `

It works well when there is at least one parameter that is set.
I tried setting an empty array but that just passes that empty array as argument to my method which requires a nullable int

**Expected behavior**
It'd be great if setInput can be nullable (since I don't always know if the AI decides to use parameters or not). 

**The Neuron AI version you are using:**
Latest

**Additional context**
Relevant stacktrace:

```
[object] (TypeError(code: 0): NeuronAI\Tools\Tool::setInputs(): Argument #1 ($inputs) must be of type array, null given, called in /Volumes/Source/leantime-oss/app/Plugins/Copilot/AiProviders/Anthropic.php on line 104 at /Volumes/Source/leantime-oss/vendor/inspector-apm/neuron-ai/src/Tools/Tool.php:95)
[stacktrace]
#0 /Volumes/Source/leantime-oss/app/Plugins/Copilot/AiProviders/Anthropic.php(104): NeuronAI\Tools\Tool->setInputs(NULL)
#1 /Volumes/Source/leantime-oss/vendor/inspector-apm/neuron-ai/src/Providers/Anthropic/HandleStream.php(65): Leantime\Plugins\Copilot\AiProviders\Anthropic->createToolMessage(Array)
#2 /Volumes/Source/leantime-oss/vendor/inspector-apm/neuron-ai/src/Agent.php(137): Leantime\Plugins\Copilot\AiProviders\Anthropic->stream(Array, Object(Closure))
#3 /Volumes/Source/leantime-oss/app/Plugins/Copilot/Services/Copilot.php(122): NeuronAI\Agent->stream(Array)
#4 /Volumes/Source/leantime-oss/app/Plugins/Copilot/Controllers/Copilot.php(83): Leantime\Plugins\Copilot\Services\Copilot->generateStreamingResponse('What are my mos...', 1, 1)
#5 /Volumes/Source/leantime-oss/app/Core/Controller/Controller.php(73): Leantime\Plugins\Copilot\Controllers\Copilot->streamResponse(Array)
```

[...]

PR #20, opened 2025-04-05
fix: Fixes input type for anthropics tool messages in cases where input is null

marcelfolaron:
When input parameters are nullable the tool message chain will send `input: null` or in cases of empty arrays `input: []`
However, Anthropic requires input to be a json object. 

In this PR I just updated the input type for the tool message so that we don't have any downstream effects on tool execution. Haven't run into any issues so far.

### DIFF
--- a/src/Providers/Anthropic/Anthropic.php
+++ b/src/Providers/Anthropic/Anthropic.php
@@ -103,9 +103,10 @@ class Anthropic implements AIProviderInterface
             ->setInputs($content['input']??[])
             ->setCallId($content['id']);
 
+        $content['input'] ??= (object)[];
+
         return new ToolCallMessage(
             [$content],
-            [$tool] // Anthropic call one tool at a time. So we pass an array with one element.
         );
     }
 }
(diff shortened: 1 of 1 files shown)


marcelfolaron:
Hey, I was able to fix #19 using a simple null check and setting the input var in the message chain to an empty object. I haven't tested the providers so this is just a simple proof of concept

ilvalerione:
You found the right fix I believe. The error happens because during serialization and deserialization of the empty inputs object PHP lost the original nature of the JSON object.

Running the code below: 

```
var_dump(json_encode(json_decode('{}', true)));
```

It will return a string '[]' instead of  the original '{}'. The cast you propose fix the error.

I also tested for other providers with no inputs and it should works either.

I'm going to close the PR because I included other changes, but the fix is correct. Thank you for the contribution.

marcelfolaron:
Thanks, glad it helped.

Just as a tip from a fellow maintainer: If the fix proposed is exactly the fix that makes it into the release it would be good to either merge the PR or at least pull in the commit/branch from the contributor so it gives credit and shows up as contribution. Just closing the PR gives no credit back to the contributor. And as it also helps your repo to have more contributors listed.

In any case, thank you for your work on this! Looking forward to diving deeper and contributing more.

ilvalerione:
Oh absolutely! This is clearly lack of experience :)

Thank you for sharing your thoughts.

PR #37, opened 2025-04-19
feat: add doctrine vector store

AlexandreCavanna:
This pull request introduces the initial implementation of a vector store integration using Doctrine based on code originally from [llphant](https://github.com/LLPhant/LLPhant).

The goal is to enable storage and retrieval of vector embeddings directly through the database layer using Doctrine entities. This is part of the broader effort to support semantic search and similarity-based operations within the application.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,6 +9,20 @@ on:
 jobs:
     ci:
         runs-on: ${{ matrix.os }}
+        services:
+            mariadb:
+                image: mariadb:11
+                env:
+                    MYSQL_DATABASE: neuron_ai_test
+                    MYSQL_USER: root
+                    MYSQL_ALLOW_EMPTY_PASSWORD: yes
+                ports:
+                    - 3306:3306
+                options: >-
+                    --health-cmd="healthcheck.sh --connect --innodb_initialized"
+                    --health-interval=10s
+                    --health-timeout=5s
+                    --health-retries=3
         strategy:
             matrix:
                 os: [ubuntu-latest]
--- a/composer.json
+++ b/composer.json
@@ -18,15 +18,15 @@
         "symfony/serializer": "^5.4 || ^6.4 || ^7.0",
         "symfony/validator": "^5.4 || ^6.4 || ^7.0",
         "symfony/property-access": "^5.4 || ^6.4 || ^7.0",
-        "phpdocumentor/reflection-docblock": "^5.3 || ^5.6",
-        "doctrine/orm": "^2.20"
+        "phpdocumentor/reflection-docblock": "^5.3 || ^5.6"
     },
     "suggest": {
         "elasticsearch/elasticsearch": "^7.0|^8.0",
         "inspector-apm/inspector-php": "^3.9",
         "symfony/process": "^4.0|^5.0|^6.0|^7.0",
         "typesense/typesense-php": "^5.0",
-        "php-http/curl-client": "^2.3"
+        "php-http/curl-client": "^2.3",
+        "doctrine/orm": "^2.20"
     },
     "require-dev": {
         "elasticsearch/elasticsearch": "^8.0",
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,9 @@
         "phpunit/phpunit": "^9.0",
         "symfony/process": "^5.0|^6.0|^7.0",
         "typesense/typesense-php": "^5.0",
-        "php-http/curl-client": "^2.3"
+        "php-http/curl-client": "^2.3",
+        "symfony/cache": "^5.4 || ^6.4 || ^7.0",
+        "ext-pdo": "*"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -14,11 +14,12 @@
         "guzzlehttp/guzzle": "^7.0",
         "psr/log": "^1.0|^2.0|^3.0",
         "psr/http-message": "^1.0|^2.0",
-        "symfony/property-info": "^6.4 || ^7.0",
-        "symfony/serializer": "^6.4 || ^7.0",
-        "symfony/validator": "^6.4 || ^7.0",
-        "symfony/property-access": "^6.4 || ^7.0",
-        "phpdocumentor/reflection-docblock": "^5.6"
+        "symfony/property-info": "^5.4 || ^6.4 || ^7.0",
+        "symfony/serializer": "^5.4 || ^6.4 || ^7.0",
+        "symfony/validator": "^5.4 || ^6.4 || ^7.0",
+        "symfony/property-access": "^5.4 || ^6.4 || ^7.0",
+        "phpdocumentor/reflection-docblock": "^5.3 || ^5.6",
+        "doctrine/orm": "^2.20"
     },
     "suggest": {
         "elasticsearch/elasticsearch": "^7.0|^8.0",
--- a/src/HandleStructured.php
+++ b/src/HandleStructured.php
@@ -21,6 +21,7 @@ use NeuronAI\StructuredOutput\Deserializer;
 use NeuronAI\StructuredOutput\JsonExtractor;
 use NeuronAI\StructuredOutput\JsonSchema;
 use Symfony\Component\Validator\ConstraintViolation;
+use Symfony\Component\Validator\Mapping\Loader\AnnotationLoader;
 use Symfony\Component\Validator\Mapping\Loader\AttributeLoader;
 use Symfony\Component\Validator\Validation;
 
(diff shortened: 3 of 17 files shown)


ilvalerione:
Such an implementation should be accompanied with tests.

AlexandreCavanna:
Absolutely!

This is a proof of concept for my work, and I’m currently juggling with my forked dev branch, which is why I haven’t run any tests yet. I plan to add unit tests following the same approach used for the other vector stores you implemented.

At this stage, I’ve managed to get the RAG working, but there are still a few things I need to adjust and refine.

ilvalerione:
> Absolutely!
> 
> This is a proof of concept for my work, and I’m currently juggling with my forked dev branch, which is why I haven’t run any tests yet. I plan to add unit tests following the same approach used for the other vector stores you implemented.
> 
> At this stage, I’ve managed to get the RAG working, but there are still a few things I need to adjust and refine.

Thank you for working on it

AlexandreCavanna:
@ilvalerione For the purpose of my POC, I needed the symfony components to remain compatible with Symfony 5.4, which will continue to receive security fixes until 2029.

During this work, I encountered an issue because Symfony 5.4 does not include the AttributeLoader class, which was only introduced in Symfony 6.4 to support PHP 8 attributes natively with Validator component.

Ensuring compatibility with Symfony 5.4 could also help promote and facilitate the adoption of the framework, especially while Symfony 5.4 remains officially supported.

I made sure that the changes stay fully compatible with both Symfony 5.4 and newer versions.
Let me know if you prefer a different approach or if I should isolate these adjustments.

ilvalerione:
I took a look at the code and I think it's ok. By the way you should move the `doctrine/orm` dependency to the suggest section.

AlexandreCavanna:
> By the way you should move the doctrine/orm dependency to the suggest section.

Done ✅

AlexandreCavanna:
There is a need to make other changes or you can plan the merge of this PR work @ilvalerione ?

ilvalerione:
There are so much Symfony specific things here. Don't you think it would be better to publish it as a Symfony bundle? We could still list it in the official documentation so it can find rapid adoption and contributions.

AlexandreCavanna:
What kind of symfony things do you think? 

If the topic is only about Doctrine, this ORM can be used as a standalone dependency outside symfony right?
But I must agree this dependency need to be use with the Symfony Cache component but you already use some component from symfony with some previous version of neuron AI and the current one like, serializer, validator, property-info etc..

Otherwise I can help to write documentation about this vector store to help users to use the framework with this implementation.

ilvalerione:
@AlexandreCavanna I will jot down a bit of documentation based on the code in the test cases. I will send you a link here so we can improve it together based on your feedback.

ilvalerione:
Why the `DoctrineEmbeddingEntityBase` is abstract? 

Can you write here the snippet of how you use this vector store in your agent with class imports?

AlexandreCavanna:
All right, I'll be glad to help!

By the way the work which is realized here will be used on my side to develop our AI Agent in the platform of reference for the communities, reinforce the democracy, facilitate your decision-making and inspire confidence by offering to the citizens the possibility of contributing actively to the public decisions.

I thought you'd be pleased to hear about our adoption of the framework haha

ilvalerione:
You should definitely post about this project in the forum so our team can spread the word!

Here is the thread to post: https://github.com/inspector-apm/neuron-ai/discussions/30

AlexandreCavanna:
> Why the `DoctrineEmbeddingEntityBase` is abstract?
> 
> Can you write here the snippet of how you use this vector store in your agent with class imports?

```
<?php

namespace Capco\AppBundle\Entity\Vector;

use Capco\AppBundle\Repository\ResponseVectorRepository;
use Doctrine\ORM\Mapping as ORM;
use NeuronAI\RAG\VectorStore\Doctrine\DoctrineEmbeddingEntityBase;

#[ORM\Entity(repositoryClass: ResponseVectorRepository::class)]
#[ORM\Table(name: 'response_vector')]
class ResponseVector extends DoctrineEmbeddingEntityBase
{
    #[ORM\Id]
    #[ORM\Column(type: 'guid', unique: true)]
    #[ORM\GeneratedValue(strategy: 'UUID')]
    public mixed $id;
}
```

He we generate embeddings from response to a questionnaire.

The DoctrineEmbeddingEntityBase class serves as a reusable base class for entities representing vector-based document embeddings. It is designed to be extended by concrete Doctrine entities that require shared fields such as embeddings, content, and metadata (e.g. sourceType, chunkNumber).

This class is not directly mapped to a database table, but rather provides mapped fields to be inherited by subclasses.

 ⚠️  If the embedding length is not 3072 you will need to specify it in the entity by overriding the $embedding property. Typically, if you use the OpenAIEmbeddingsProvider class or an other embeddings provider with a dimension of 1536, you will need to set the length to 1536 in the entity.

ilvalerione:
So it can be a normal class instead of abstract. If developers wat to extend it to change the ID type they can do it anyway.

AlexandreCavanna:
Yes indeed

ilvalerione:
Ok, can you send me a snippet of how to construct the DoctrineVectorStore in the agent class?

AlexandreCavanna:
```
// Agent Class
<?php

namespace Capco\AppBundle\Agent;

use Capco\AppBundle\Entity\Vector\ResponseVector;
use Doctrine\ORM\EntityManagerInterface;
use NeuronAI\Agent;
use NeuronAI\RAG\VectorStore\Doctrine\DoctrineVectorStore;
use NeuronAI\RAG\VectorStore\VectorStoreInterface;

class CapcoAgent extends Agent
{
    public function __construct(
        private readonly EntityManagerInterface $entityManager
    ) {
    }

    public function vectorStore(): VectorStoreInterface
    {
        return new DoctrineVectorStore($this->entityManager, ResponseVector::class);
    }
}

    
// Client code
CapcoAgent::make($this->entityManager)  
```

ilvalerione:
$this->entity Manager? Where do you get it?

AlexandreCavanna:
This is a symfony thing but without symfony, you need to setup database configuration like this.

```
<?php
require_once "vendor/autoload.php";

use Doctrine\DBAL\DriverManager;
use Doctrine\ORM\EntityManager;
use Doctrine\ORM\ORMSetup;

$paths = ['/path/to/entity-files'];
$isDevMode = false;

$dbParams = [
    'driver'   => 'pdo_mysql',
    'user'     => 'root',
    'password' => '',
    'dbname'   => 'foo',
];

$config = ORMSetup::createAttributeMetadataConfiguration($paths, $isDevMode);
$connection = DriverManager::getConnection($dbParams, $config);
$entityManager = new EntityManager($connection, $config);
```

https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/configuration.html#obtaining-an-entitymanager

$entityManager can be use after that.

ilvalerione:
Do you have a snippet on how to get the entity manager in Symfony? Maybe it can be gathered from the container, and then the link to doctrine can be mentioned as instructions for other environments.

AlexandreCavanna:
With Symfony there is nothing to do more than to make sure your class is declare as a service and you can inject directly entityManager in your constructor like this:

```

// config/services.yaml
services:
  _defaults:
    autowire: true
    autoconfigure: true

    App\:
    resource: '../src/'

// src/CapcoAgent.php
    public function __construct(
        private readonly EntityManagerInterface $entityManager
    ) {
    }
```

ilvalerione:
Here is the final doc: 
https://docs.neuron-ai.dev/components/vector-store#doctrine-orm

Feel free to give us your feedback if it can be improved.

The Doctrine vector store was released with [1.10.0](https://github.com/inspector-apm/neuron-ai/releases/tag/1.10.0)

AlexandreCavanna:
It seems correct to me this covers the minimum required for it to work. I’m looking forward to getting feedback from other contributors on this implementation in the future!

It’s working on my side, and we’ve validated our POC with Neuron AI. I’ve encouraged our product team to contribute to this project as well.

AlexandreCavanna:
I spoke too fast haha

`Doctrine\ORM\EntityManager is declared as a service` in fact this is not `EntityManager` but `MyChatBot`

AlexandreCavanna:
This part of documentation should be edit in this way:

If you are using a Symfony application, there’s nothing else you need to configure regarding the Entity Manager — Doctrine\ORM\EntityManagerInterface is already available as a service by default.

However, you must declare your RAG agent class (e.g., MyChatBot) as a service in your services.yaml file so that Symfony can inject the Entity Manager properly into it via autowiring.

In the code snippet, we type-hint against the interface (EntityManagerInterface) rather than the concrete class. Symfony is service container automatically resolves it to the default Doctrine\ORM\EntityManager.

```
use Doctrine\ORM\EntityManagerInterface;
use NeuronAI\RAG\RAG;
use NeuronAI\RAG\VectorStore\Doctrine\DoctrineEmbeddingEntityBase;
use NeuronAI\RAG\VectorStore\VectorStoreInterface;

class MyChatBot extends RAG
{
    public function __constructor(
        private readonly EntityManagerInterface $entityManager
    ){}
    
    {
        return new DoctrineVectorStore(
            entityManager: $this->entityManager,
            entityClassName: DoctrineEmbeddingEntityBase::class
        );
    }
}
```

ilvalerione:
😅 thank you for the feedback.

I fixed the doc based on your example.

Issue #39, opened 2025-04-19
Duplicated messages in chat history


iztok:
**Describe the bug**
When using tools, history contains duplicated messages of the assistant's last response. 

**To Reproduce**
1. Create agent with chatHistory()
2. Add tools
3. Make the agent use a tool
4. Check history and see the last message is duplicated
5. Bonus: Some messages have content "null". This breaks OpenAi's API when using this message in the request.

**Expected behavior**
There should be only one last message

**Screenshots**
/

**The Neuron AI version you are using:**
 - 

**Additional context**

Agent
```
<?php

namespace App\Agents;

use NeuronAI\Agent;
use App\Models\User;
use NeuronAI\Tools\Tool;
use NeuronAI\Tools\ToolProperty;
use NeuronAI\Providers\OpenAI\OpenAI;
use NeuronAI\Chat\History\FileChatHistory;
use NeuronAI\Providers\AIProviderInterface;
use NeuronAI\Chat\History\AbstractChatHistory;
use NeuronAI\Chat\History\InMemoryChatHistory;

class TestAgent extends Agent
{
    protected function provider(): AIProviderInterface
    {
        return new OpenAI(
            key: config('services.openai.api_key'),
            model: 'gpt-4o-mini'
        );
    }

    protected function chatHistory(): AbstractChatHistory
    {
        // return new InMemoryChatHistory(
        //     contextWindow: 50000
        // );
        return new FileChatHistory(
            directory: storage_path(),
            key: 'test-agent',
            contextWindow: 50000
        );
    }

    protected function tools(): array
    {
        return [
            Tool::make(
                'get_user_info',
                'Retrieve the user info.',
            )->addProperty(
                new ToolProperty(
                    name: 'user_id',
                    type: 'string',
                    description: 'The ID of the user.',
                    required: true
                )
            )->setCallable(function (string $user_id) {
                $user = User::find($user_id);
                if (!$user) {
                    return ['error' => 'User not found', 'requested_id' => $user_id];
                }
                return json_encode($user);
            })
        ];
    }
}
```

Usage:
```
    $agent = new TestAgent();
    $response = $agent->chat(
        new UserMessage("Get user info for user 1")
    );
```

History file:
```
[
    {
        "role": "user",
        "content": "Get user info for user 1"
    },
    {
        "tool_calls": [
            {
                "id": "call_R47wL5p1pVsmo65Zzy7t4lS3",
                "type": "function",
                "function": {
[...]

PR #41, opened 2025-04-19
Fix/duplicated messages in chat history

iztok:
Related to #39


ilvalerione:
What an error man! Thank you!

PR #61, opened 2025-05-04
Uploading Images to Anthropic

JanneDeVos:
Implementation for uploading images. This Pull Request has an implementation for Anthropic, but the UserImage class format should also be able to function for other AI providers

### DIFF
--- a/src/Chat/Messages/Image.php
+++ b/src/Chat/Messages/Image.php
@@ -2,10 +2,10 @@
 
 namespace NeuronAI\Chat\Messages;
 
-class UserImage extends Message
+class Image
 {
     public function __construct(public string $image, public string $type = 'url', public ?string $mediaType = null)
     {
-        parent::__construct(Message::ROLE_USER);
+        //
     }
 }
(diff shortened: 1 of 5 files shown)


ilvalerione:
As mentioned in the Anthropic documentation a message can have role, content "and/or" an image attached.

By the way it's basically the same for other LLMs.

So the UserImage message doesn't really represent a right API for this use case. I believe image should be an attachment to the current message class.

If you want to move forward with this implementation consider to also develop some unit tests.

JanneDeVos:
Since you’re the maintainer, I think you should decide on the final implementation. Is this the implementation you would prefer? I will do the refactor if so. Will also write some tests in the process.
```
new UserMessage('This is my content!', new Image($properties));
```

JanneDeVos:
So know we can add images to the UserMessage like this:
```
$response = $agent->structured([
    (new UserMessage('Content'))
        ->addImage(new Image($imageData, 'base64', 'image/png')),
]);
```

ilvalerione:
Here is the final implementation supported by all AI providers in the framework (Anthropic, Gemini, OpenAI, Ollama):

https://docs.neuron-ai.dev/attach-images

PR #68, opened 2025-05-06
Adding ArrayOf attribute for arrays in Structured Output

JanneDeVos:
While trying Structured Output I tried using a list with the doc-block like this:
```
/**
* @var array<\App\Agent\Models\Tag>
*/
```
I could not get it to work. The problem was I did a comment with an imported class like this:
```
/**
* @var array<Product>
*/
```
And my Phpstorm really wants me to use an import. So thats why I created the ArrayOf attribute. So now it's possible to do this:
```
#[ArrayOf(Product::class)]
public array $products;
```

### DIFF
--- a/src/StructuredOutput/ArrayOf.php
+++ b/src/StructuredOutput/ArrayOf.php
@@ -1,0 +1,12 @@
+<?php
+
+namespace NeuronAI\StructuredOutput;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
+class ArrayOf
+{
+    public function __construct(public string $class)
+    {}
+}
--- a/src/StructuredOutput/ArrayOf.php
+++ b/src/StructuredOutput/ArrayOf.php
@@ -3,10 +3,15 @@
 namespace NeuronAI\StructuredOutput;
 
 use Attribute;
+use InvalidArgumentException;
 
 #[Attribute(Attribute::TARGET_PROPERTY)]
 class ArrayOf
 {
     public function __construct(public string $class)
-    {}
+    {
+        if (!class_exists($class) && !enum_exists($class)) {
+            throw new InvalidArgumentException("Class $class does not exist.");
+        }
+    }
 }
--- a/src/StructuredOutput/JsonSchema.php
+++ b/src/StructuredOutput/JsonSchema.php
@@ -163,6 +163,7 @@ class JsonSchema
 
             // Parse PHPDoc for the array item type
             $docComment = $property->getDocComment();
+            $attributes = $property->getAttributes(ArrayOf::class);
             if ($docComment) {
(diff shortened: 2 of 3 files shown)


ilvalerione:
It doesn't matter that you imported the class. You can always declare the full path in the comment:

```php
use App\Agent\Models\Tag


/**
* @var array<\App\Agent\Models\Tag>
*/
```

JanneDeVos:
So this is what my Phpstorm displays. I think some aggressive code analysers will probably remove this doc:
Scherm­afbeelding 2025-05-12 om 20 21 34

And this what I think is cleaner than the full import isn't working:
```
/**
* @var array<Product>
*/
```

ilvalerione:
Since it doesn't involve an error but just an IDE alert I prefer to avoid custom implementation out of the Symfony validation rules.

JanneDeVos:
To be honest, I’m not familiar with Symfony myself—I work almost exclusively with Laravel. I’m of course familiar with the `@var array<Product>` syntax from docblocks, but I’m used to always being able to use imports for type references like this.

The implementation in my PR is based on a widely-used Laravel package: [Laravel Data](https://spatie.be/docs/laravel-data) by Spatie. As you can see in [their documentation](https://spatie.be/docs/laravel-data/v4/as-a-data-transfer-object/nesting), they support multiple syntaxes for type declarations. I personally prefer the latter, where classes are explicitly imported.

ilvalerione:
It's not about Symfony or Laravel here. It's just an alert on the IDE. Since it doesn't fix errors, and does not introduce features I'm going to close the PR.

JanneDeVos:
I’m a bit surprised to see this PR being closed, as the proposed change is purely additive and fully opt-in. It doesn’t alter existing behavior or introduce any side effects—rather, it provides improved structure and better tooling support for those who wish to use it.

In my view, using #[ArrayOf(...)] is more aligned with the modern direction PHP is taking, especially considering your own use of attributes like #[SchemaProperty()] or validation attributes like #[NotBlank]. Compared to those, docblock annotations start to feel a bit out of place in an otherwise attribute-driven design.

ilvalerione:
The problem here is the implementation itself. You are extending the conversion into the json schema that it's  not the point of attributes. Attributes are for data validation. Since it is implemented using the Symfony Validator your attribute has no effect on it. 

I would like to embrace this PR if it is the start of an internal validation system driven by attributes to entirely eliminate the Symfony dependency since I don't really like because of all external dependencies it brings into the project. And start the Neuron way to accomplish this task.

It requires to create an attribute interface with the "validate" method and a Validator class that performs validation for all compatible attributes. The validator should collect all exception messages generated by the not-valid attributes and return a violation collections. 

This could be a great contribution.

ilvalerione:
For a more extensive development on a new internal validator component we will work on a new branch.

PR #72, opened 2025-05-07
Adding static Message creation

JanneDeVos:
Agents have a static constructor. We can use Agent::make() to instantiate a new class.

After implementing image support, I noticed the syntax used in UserMessage with an image — and it hurt my eyes:
```
$message = (new UserMessage("Describe this image"))
    ->addImage(new Image(
        image: 'https://url_of/image.jpg', 
        type: Image::TYPE_URL
    ));
```
Starting from PHP 8.4, we can omit the outer parentheses, but I was still looking for a cleaner solution.

With the CallStatic trait, we can now simply write:
```
$message = Message::UserMessage("Describe this image")
    ->addImage(Image::make('https://url_of/image.jpg'));
```

### DIFF
--- a/src/Chat/Messages/CallStatic.php
+++ b/src/Chat/Messages/CallStatic.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace NeuronAI\Chat\Messages;
+
+use NeuronAI\StaticConstructor;
+
+trait CallStatic
+{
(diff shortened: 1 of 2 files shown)


ilvalerione:
I don't like this kind of patterns because it makes maintainability harder.

PR #78, opened 2025-05-12
Adding support for attaching documents like pdfs to chats

JanneDeVos:
It is currently possible to add images to chats. This PR makes it possible to also add documents (depending on the AI provider).

To achieve this, I refactored the existing image support into a more abstract Attachment parent class. The Attachment class also makes it easier to support other file types in the future, such as audio or video.

It is now possible to add PDF documents to a chat, either as a URL or as base64-encoded content.


ilvalerione:
Can you add a description of this implementation, the intended use cases and the design choices?

nunodonato:
@JanneDeVos this is awesome! Not being able to include pdfs is a deal-breaker for me, its so much better to use pdfs directly than having to extract text first.

ilvalerione:
@JanneDeVos we are in the process of changing all the "const" with explicit enum classes. Since it is a braking change I'm going to merge this great implementation but I will release a bit later to implement enums also for this changes. It should result in a very complete development. 

Thank you for working on it.

ilvalerione:
Here is the final implementation:

https://docs.neuron-ai.dev/attachments-documents-and-images

Released with [1.10.0](https://github.com/inspector-apm/neuron-ai/releases/tag/1.10.0)

PR #84, opened 2025-05-15
Implemented CohereRerankerPostProcessor

asterixcapri:
Added Cohere reranker integration as an additional option for document reranking in RAG pipelines. Cohere's solution is one of the most widely used reranking services in production RAG systems, making it a valuable addition to the framework.


ilvalerione:
Here is the documentation for the latest components: https://docs.neuron-ai.dev/post-processor#cohere-reranker

Issue #87, opened 2025-05-15
When a tool from a MCP server does not return the 'description' field we are getting an error


RazielRodrigues:
When a tool from a MCP server does not return the 'description' field for the tool we are getting an error inside the class McpConnector in the function createTool line 74.

I bet it is because is expecting an string but since the key does not exists from the tool side it is returning NULL and it leads to an error.

![Image](https://github.com/user-attachments/assets/b95599e6-50eb-461b-972e-53fe8292c959)

I don't know if it has a standard for the tools struct when comes from a MCP Server, so in this case I won't open this as a bug for Neuron but just to let you guys know that I found this and I was not able to proceed with the development until I added a workaround in the line 74 to pass the loop I have done a simple nullish operator returning an empty string, something like: `$input['description'] ?? ''"`

How I have done the debug:

![Image](https://github.com/user-attachments/assets/15ad7101-60ba-410a-9862-fffcbd2c5b6c)

this is the result from the debug:

```
array(3) {
  ["description"]=>
  string(109) "Scrape search results from Google, Bing or Yandex. Returns SERP results in markdown (URL, title, description)"
  ["inputSchema"]=>
  array(5) {
    ["type"]=>
    string(6) "object"
    ["properties"]=>
    array(2) {
      ["query"]=>
      array(1) {
        ["type"]=>
        string(6) "string"
      }
      ["engine"]=>
      array(3) {
        ["type"]=>
        string(6) "string"
        ["enum"]=>
        array(3) {
          [0]=>
          string(6) "google"
          [1]=>
          string(4) "bing"
          [2]=>
          string(6) "yandex"
        }
        ["default"]=>
        string(6) "google"
      }
    }
        }
        ["default"]=>
        string(6) "google"
      }
    }
        ["default"]=>
        string(6) "google"
      }
    }
        string(6) "google"
      }
    }
      }
    }
    ["required"]=>
    }
    ["required"]=>
    ["required"]=>
    array(1) {
      [0]=>
      string(5) "query"
    }
    ["additionalProperties"]=>
    bool(false)
    ["$schema"]=>
    array(1) {
      [0]=>
      string(5) "query"
    }
    ["additionalProperties"]=>
    bool(false)
    ["$schema"]=>
    string(39) "http://json-schema.org/draft-07/schema#"
  }
  ["name"]=>
  string(13) "search_engine"
}
    array(1) {
      [0]=>
      string(5) "query"
    }
    ["additionalProperties"]=>
[...]

PR #99, opened 2025-05-22
fix: add nullish operator when tools does not have description

RazielRodrigues:
Hello!

I mentioned this bug: https://github.com/inspector-apm/neuron-ai/issues/87

and I have seen the fix from you but it persisted in my local tests, so in order to solve it I have opened this MR with the fix that worked for me, please let me know if I need to do something else to help the project.

stacktrace even after upgrading the project

```
 PHP Warning:  Undefined array key "description" in \vendor\inspector-apm\neuron-ai\src\MCP\McpConnector.php on line 64
PHP Fatal error:  Uncaught TypeError: NeuronAI\Tools\ToolProperty::__construct(): Argument #3 ($description) must be of type string, null given, called in \vendor\inspector-apm\neuron-ai\src\MCP\McpConnector.php on line 61 and defined in \vendor\inspector-apm\neuron-ai\src\Tools\ToolProperty.php:7
```

### DIFF
--- a/src/MCP/McpConnector.php
+++ b/src/MCP/McpConnector.php
@@ -40,7 +40,7 @@ class McpConnector
     {
         $tool = Tool::make(
             name: $item['name'],
-            description: $item['description']??''
+            description: $item['description'] ?? ''
         )->setCallable(function (...$args) use ($item) {
             $response = call_user_func([$this->client, 'callTool'], $item['name'], $args);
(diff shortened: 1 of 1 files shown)


ilvalerione:
Ah ok, I added the fallback only for tool's description, but the tool property also needed to be fixed.

Thank you for the contribution.

PR #102, opened 2025-05-22
WIP - Static StateGraph implementation

sixty-nine:
This PR implements a static StateGraph of agents. 

See the [full story](https://github.com/inspector-apm/neuron-ai/discussions/47), and the [comments on the proof-of-concept](https://github.com/sixty-nine/neuron-ai/pull/1).

This PR is open to have a base for discussion but it needs more work and polishing. Notably, the `src/Workflow/WorkflowAgent.php` currently only implements the `chat` method.

Example of usage: 

```php
$geocodeTool = Tool::make('get_timezone', 'Get the timezone of a location.')
    ->addProperty(
        new ToolProperty(
            name: 'location',
            type: 'string',
            description: 'The location to get the timezone of.',
            required: true,
        )
    )
    ->setCallable(fn (string $location) => 'CET');

$timeTool = Tool::make('get_time_and_date', 'Get the current time and date.')
    ->addProperty(
        new ToolProperty(
            name: 'timezone',
            type: 'string',
            description: 'The timezone.',
            required: true,
        )
    )
    ->setCallable(
        fn(string $timezone) => (new \DateTimeImmutable("now", new \DateTimeZone($timezone)))->format('Y-m-d H:i:s')
    );

$agent1 = Agent::make()
    ->withProvider($provider)
    ->withInstructions('You are an AI agent specialized in retrieving the timezone for a location')
    ->addTool($geocodeTool);

$agent2 = Agent::make()
    ->withProvider($provider)
    ->withInstructions('You are an AI agent specialized in giving the current time')
    ->addTool($timeTool);

$graph = (new StateGraph())
    ->addNode('a', $agent1)
    ->addNode('b', $agent2)
    ->addEdge(StateGraph::START_NODE, 'a')
    ->addEdge('a', 'b')
    ->addEdge('b', StateGraph::END_NODE);

$reply = WorkflowAgent::make($graph)->chat(new UserMessage('What time is it in Paris/France'));
```

PR #109, opened 2025-05-24
Fix for Gemini response when called in workflow

peter-mw:
Fix of

Undefined array key "candidatesTokenCount" in this file vendor\inspector-apm\neuron-ai\src\Providers\Gemini\HandleChat.php

Seems when calling Gemini in a workflow defined in https://github.com/inspector-apm/neuron-ai/pull/102 
  it does not include candidatesTokenCount every time 

image


ilvalerione:
Thank you

PR #134, opened 2025-05-31
refactor platform check to allow doctrine/dbal:^4

tacman:
I wish I know CI better, so we could check all the versions of doctrine during testing.

I believe, though, that this will work for all, as the current code fails with dbal 4.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@ name: Tests
 
 on:
     push:
-        branches: [ "main" ]
+        branches: [ "main","tac" ]
     pull_request:
         branches: [ "main" ]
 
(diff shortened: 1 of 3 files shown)


tacman:
oops, I didn't mean to check everything in, just the src changes.

tacman:
Thanks.  There are still a few instances of the old way of calling NotSupported.  Fixing  #136 should take care of those.

ilvalerione:
Is this related?

https://github.com/inspector-apm/neuron-ai/actions/runs/15375826569/job/43260431226#step:8:9

tacman:
I hadn't intended to push that.  I don't know CI well enough, it was a test to see if the system would show any more errors with dbal/4.  It should be removed, and probably we have have tests for both lowest and highest dependencies.

I'm hoping that 

```
symfony new --webapp my-ai-test && cd my-ai-test
composer require (neuron-ai)
(add agents and loaders, etc.)
composer outdated
```

doesn't have any problems or require reverting to an older library.

PR #138, opened 2025-06-01
Move AzureOpenAi provider to correct folder

praczynski:


ilvalerione:
Now I see it was a missmatch with the namespace. Thank you

PR #147, opened 2025-06-04
Fix json with incorrectly formatted structure

mappweb:
Fix json with incorrectly formatted structure for when it has no parameters


ilvalerione:
Can you describe the behavior of OpenAI in this case? Did you get an exception?

ilvalerione:
In our tests it doesn't cause any error.

PR #164, opened 2025-06-14
RAG Custom Document Model

ilvalerione:
This pull request aim to add support for custom `Document` model class in RAG implementations. You can extend the default Document class to add new fields in the form of a DTO:

```php
class CustomDocumentModel extends Document
{
    public string $user_id;
}
```

Once you create your own document model you have to declare this custom class in the RAG implementation:

class RagAgent extends RAG
{
    /**
     * The default document model.
     */
    protected string $documentModel = CustomDocumentModel::class;
}

All the additional fields you add to the custom document model will be mapped in the vector store automatically, but you have to fill the additional property with values:

```php
$documents = StringDataLoader::for($text)
    ->withDocumentModel(CustomDocumentModel::class) // <- Tell the data loader about the custom document model
    ->getDocuments();

$documents[0]->user_id = 23

RagAgent::make()->addDocuments($documents);
```

## Support for vector store hybrid search

Some vector stores support hybrid search that means the ability to limit the similarity search to a sub-set of records that is created adding filters on document fields. 

Since you have now the ability to store custom fields into the vector store you can eventually use hybrid search to make your RAG more accurate.

```php
class RagAgent extends RAG
{
    protected array $pineconeFilters = [];

    ...
    
    {
        return (new PineconeVectoreStore(
            key: 'PINECONE_API_KEY',
            indexUrl: 'PINECONE_INDEX_URL'
        ))->withFitlers($this->pineconeFilters);
    }

    public function withPineconeFilters(array $filters): self
    {
        $this->pineconeFilters = $filters;
        return $this;
    }
}
```

With this implementation you have two ways to pass the filters:

```php
// Pass filters at runtime
RagAgent::make()->withPineconeFilters([
    'user_id' => ['$eq' => 23]
])->answer( new UserMessage(...));
```

> By default this implementation shouldn't break the actual flow, rather than the removal of two fields from the default `Document` model `chunkNumber` and `hash`. If you think these fields were important they can be restored in the default model.


asterixcapri:
Ciao Valerio, really interesting PR!

To be honest, I was expecting a lighter and less invasive approach. I had been experimenting with simply adding a `metadata` field to the `Document`, similar to what LangChain or LlamaIndex do. Probably for many common use cases (like `lang`, `url`), that seemed sufficient and easy to use without introducing new abstractions.

That said, the `CustomDocumentModelInterface` approach is definitely more solid, especially if you want to represent actual domain objects or handle more complex logic. I totally see the tradeoff between simplicity and stronger structure.

In the end, I think your proposal makes more sense in terms of type safety and flexibility, especially as the library grows.

If anyone else in the community has other thoughts or needs, it would be great to hear them too.

ilvalerione:
Yes it was my doubt during implementation, a bit invasive. So basically the Document can implement get/set metadata that can be also used to dynamically map custom fields into the vector store.

PR #169, opened 2025-06-16
Document metadata

ilvalerione:
This PR aims to add support for Document metadata in RAG implementations. You can attach some custom metadata to the default Document instance that will be automatically saved under the "metadata" field in the vector store:

```php
$documents = StringDataLoader::for($text)
    ->getDocuments();

foreach($documents as $document) {
    $document->addMetadata('user_id', 1234);
}

RagAgent::make()->addDocuments($documents);
```

## Support for vector store hybrid search
Some vector stores support hybrid search that means the ability to limit the similarity search to a sub-set of records that is created adding filters on document fields.

Since you have now the ability to store custom metadata into the vector store you can eventually use hybrid search to narrow the scope of the search your document store.

```php
class RagAgent extends RAG
{
    protected array $pineconeFilters = [];

    ...
    
    {
        return (new PineconeVectoreStore(
            key: 'PINECONE_API_KEY',
            indexUrl: 'PINECONE_INDEX_URL'
        ))->withFitlers($this->pineconeFilters);
    }

    public function withPineconeFilters(array $filters): self
    {
        $this->pineconeFilters = $filters;
        return $this;
    }
}
```

With this implementation you can pass additional filters at runtime:

```php
// Pass filters at runtime
RagAgent::make()->withPineconeFilters([
    'user_id' => ['$eq' => 1234]
])->answer( new UserMessage(...));
```


ilvalerione:
@asterixcapri this implementatino just add the metadata field to the Document class, and use it in the vector stores to add and retrieve additional data dynamically.

asterixcapri:
I personally prefer this over the previous approach. I think it's cleaner and less invasive, while still providing the flexibility needed to handle metadata across different vector stores.

Thanks for iterating on this.

PR #178, opened 2025-06-17
Structured output enum support

mchasleslr:
# Pull Request Template

## ⚠️ Branch Requirement

- [x] I confirm this PR targets the `develop` branch

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

### What does this PR do?
This PR introduces comprehensive support for backed enums in the structured output system. It adds both deserialization capabilities for backed enums and a new `Enum` validation rule that can enforce strict value selection from predefined sets or backed enum cases.

This addresses a common use case where developers need to constrain field values to specific options instead of receiving similar but incorrect variations. The current system allowed the AI to return approximate matches rather than exact values from a defined set.

### How does this change should be used and documented?
Developers can now use backed enums to define strict value sets for their structured output properties. The `Enum` validation rule can be applied in two ways (mutually exclusive):
- With a static array of allowed values via the `values` parameter
- With a `BackedEnum` class via the `class` parameter for automatic value extraction

Example usage:
```php
enum SubjectType: string {
    case NATURAL_PERSON = 'natural person';
    case SOLE_PROPRIETORSHIP = 'sole proprietorship';
    case PARTNERSHIP = 'partnership';
    case CORPORATION = 'corporation';
}

class Registry
{
    // Implicit definition + validation (recommended)
    #[SchemaProperty(description: 'type of Subject', required: true)]
    #[Enum(class: SubjectType::class)]
    public SubjectType $subjectType;

    // Explicit definition + validation
    #[SchemaProperty(description: 'Language, should be exactly one of those : "en", "fr", "es", "it"', required: true)]
    #[Enum(values: ['en', 'fr', 'es', 'it'])]
    public string $language;
}

$registry = $agent->structured(new UserMessage("Hi!", Registry::class));

echo $registry->subjectType->value . PHP_EOL . $registry->language;
```

> [!NOTE]
>The implicit approach (using backed enums) is recommended as it provides better schema guidance to the LLM, reducing the likelihood of validation failures. The explicit approach (listing values in descriptions) has higher error rates since the LLM may return similar but incorrect values, requiring validation retries that consume additional tokens.

## Related Issues

This PR addresses some points raised in the discussion: [Strict list of values in StructuredOutput](https://github.com/inspector-apm/neuron-ai/discussions/160)

## Testing

- [x] All existing tests pass

## Breaking Changes

- [x] No breaking changes

## Documentation

- [x] New documentation needs to be created

## Checklist

## Additional Notes

Please feel free to provide feedback or suggestions on the implementation or associated documentation.


ilvalerione:
I think better APIs could be naming the validation Rule "Enum" with the properies "class" and "values".

mchasleslr:
> I think better APIs could be naming the validation Rule "Enum" with the properies "class" and "values".

Done!

PR #190, opened 2025-06-23
Do not override dynamically injected instructions

chilladelia:
When instructions are set on the agent, the tool bootstrapping process must not ignore them.


ilvalerione:
Thank you for the fix!

PR #207, opened 2025-06-30
fix postgres schema tool placeholders

simPod:
I'm using PG 16, PDO php 8.3.

This variant works and returns result. When using $1, $2... etc. queries return no results from db.


ilvalerione:
Thank you for the fix! Please send future PRs against the develop branch.

PR #210, opened 2025-07-01
make openai key optional since it is not always required

simPod:
BC break

### DIFF
--- a/src/Providers/OpenAI/OpenAI.php
+++ b/src/Providers/OpenAI/OpenAI.php
@@ -45,8 +45,8 @@ class OpenAI implements AIProviderInterface
     protected MessageMapperInterface $messageMapper;
 
     public function __construct(
-        protected string $key,
         protected string $model,
+        protected string|null $key = null,
         protected array $parameters = [],
     ) {
         $this->client = new Client([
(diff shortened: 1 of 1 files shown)


ilvalerione:
when it could be null? Maybe in openAI compatible providers? Can you add more details or examples? 

Can you send the PR against the develop branch?

Thank you for your contribution.

simPod:
We're hosting our private models on k8s cluster and it does not require any auth via header.

simPod:
Currently, I'm setting key as `''` and since Auth header is a noop it works. But not having to specify at all it seems better to me.

ilvalerione:
What is the tool you are using for private models hosting? Isn't it Ollama?

simPod:
we had ollama, now we have openai api with hosted gemma, qwen, mistral etc.
models

ilvalerione:
What is openai hosted?

simPod:
Ok so it's still Ollama with vLLM, the guy said.

ilvalerione:
You can use Ollama "or" vLLM not both...

simPod:
Ok, did not study it thoroughly, there's bazillion of other stuff running there in order for it to work.

simPod:
makes no sense?

ilvalerione:
Maybe vLLM has an OpenAI compatible API. You can just extend the OpenAI provider and pass an empty string in the constructor, overriding the baseUri property with the url of the vLLM server.

Take a look at the Grok provider to get some inspiration.

For Ollama instead there is the dedicated Ollama provider in Neuron. 

Hope it can help.

simPod:
I'm doing it like that right now. It was only a suggestion that key should
not be a required arg. Kk nvm, thanks.

PR #220, opened 2025-07-04
Extends the assignment of a callable to a tool.

mappweb:


ilvalerione:
need more context, and use cases.

mappweb:
> need more context, and use cases.

@ilvalerione  Example:

/**
* Definitión fuction calling
*/
class SumTool extends Tool
{
  public function properties()
    {
        return [
          new ToolProperty(
                name: 'a',
                type: PropertyType::INTEGER,
                description: 'Variable a',
                required: true
            ),
           new ToolProperty(
                name: 'b',
                type: PropertyType::INTEGER,
                description: 'Variable b',
                required: true
            )
      ];
    }
    public function callback()
    {
        return SumAction(SearchProductAction::class);
    }
}

/**
* Action fuction calling
*/
class SumAction
{
    public function __constructor(protected MathProvider $math){}
    
    public function __invoke(int $a, int $b)
    {
        return $this->math->sum($a, $b);
    }
}

ilvalerione:
What prevent you to do the same with actual implementation?

mappweb:
> What prevent you to do the same with actual implementation?

Separation of responsibilities between the function definition and the action to create cleaner wrapper (php artisan make:tool SumTool -a). a = SumAction. I am also building a sytem prompt hub, which will take advantage of this functionality.

ilvalerione:
You are not forced to use the __invoke method as a callable. It's just a convention. You can call the `setCallable()` method to set your callable as per your convenience. Using the built-in setCallable allows you to set the action as a callable.

mappweb:
It should be noted that it extends the use of the callable both at the class level and at the external class level. But not a matter of particular convenience.

Example 1:
class SumTool extends Tool
{
    public function callback()
    {
      return function ($a, $b,.....) {};
    }
}

Example 2: Compatibility
class SumTool extends Tool
{
    public function __invoke($a, $b)
    {
      //TODO
    }
}

Example 3: External class
class SumTool extends Tool
{
    public function callback()
    {
      return app(SumAction::class);
    }
}

Example 4:
class SumTool extends Tool
{
    public function callback()
    {
      return fn($a) => $str . $a;
    }
}

ilvalerione:
What is the structure of the SumAction class?

PR #225, opened 2025-07-10
Add Metadata to Message Class for Displaying Actual User Input

jrking4:
Currently, when working with chat histories, user requests are sometimes embedded within larger prompt templates. For example:

```
You are a helpful agent. Please respond to the user request
** User Request **
{{ $UserRequest }}

Please respond only with the answer, no other text is required
```

In these cases, retrieving the original user message for display or logging purposes requires parsing out the user's text from the overall prompt, which is error-prone and cumbersome. I also see other cases where it maybe useful to get/set metadata to messages.

This PR updates the `Message` class to support fetching metadata, it is a protected property.

### DIFF
--- a/src/Chat/Messages/Message.php
+++ b/src/Chat/Messages/Message.php
@@ -77,6 +77,14 @@ class Message implements \JsonSerializable
         return $this;
     }
 
+    public function getMetadata(): array {
+        return $this->meta;
+    }
+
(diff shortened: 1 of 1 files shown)


ilvalerione:
Thank you for the proposal. What about the other use cases? The more you can extend the scenario the better.

ilvalerione:
I believe this is more an application problem rather than a lack of Neuron.

Issue #242, opened 2025-07-28
Error when using with Anonoyms class and OpenAi Provider


peter-mw:
**Describe the bug**
When using with anonymous class i get error 

**To Reproduce**
1. Implement this 

```php
     $class = new class {
                            #[SchemaProperty(description: 'The name of the testimonial.', required: true)]
                            public string $name;

                            #[SchemaProperty(description: 'The content of the testimonial.', required: true)]
                            public string $content;


                            #[SchemaProperty(description:'The company of the client.', required: true)]
                            public string $client_company;

                            #[SchemaProperty(description: 'The role of the client.', required: true)]
                            public string $client_role;
                        };

   $resp = $agent->structured(
                                new UserMessage($prompt)
                                , $class::class,
                                maxRetries: 3

                            );
```
2. See error

```
Impossible to generate a structured response for the class class@anonymous: { "error": { "message": "Invalid 'response_format.json_schema.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.", "type": "invalid_request_error", "param": "response_format.json_schema.name", "code": "invalid_value" } }
```

**Expected behavior**

**Screenshots**

<img width="1021" height="556" alt="Image" src="https://github.com/user-attachments/assets/b0997f22-afa4-43e3-b0ba-5f17d7dc1b74" />

array:7 [▼ // vendor\inspector-apm\neuron-ai\src\Providers\OpenAI\HandleStructured.php:19
  0 => "class@anonymous\x00H:"
  1 => "debug"
  2 => "microweber"
  3 => "Modules"
  4 => "Testimonials"
  5 => "Filament"
  6 => "TestimonialsTableList.php:135$8c5"
]

Will send PR to fix shortly

ilvalerione:
PR #243  released with 1.16.5

Thank you for the contribution!

PR #243, opened 2025-07-28
Fix of #242

peter-mw:
Fix of OpenAI driver when used with anonymous class


ilvalerione:
You should send the PR against the `develop` branch,

peter-mw:
ok changed to dev branch

peter-mw:
thanks

Issue #262, opened 2025-08-14
Document ID format compliance for the Qdrant vector store


nextmwa:
**Configuration:**
Qdrant Vector Store
version: 1.14.1
commit: 530430fac2a3ca872504f276d2c91a5c91f43fa0

Neuron-ai
version: 1.16.7

**Error**:
Format error in JSON body: value 689da81268df8 is not a valid point ID

When generating a document to be sent to the Qdrant vector store, the [Document](https://github.com/inspector-apm/neuron-ai/blob/1.x/src/RAG/Document.php) class in its constructor generates a document->id using uniqid(). However, according to Qdrant’s [documentation](https://qdrant.tech/documentation/concepts/points/#point-ids), the accepted formats for point IDs are only 64-bit unsigned integers and UUIDs in various formats.
Locally, I modified the Document class as follows:

```
namespace NeuronAI\RAG;
use Ramsey\Uuid\Uuid;
public function __construct(
    public string $content = '',
) {
    $this->id = Uuid::uuid4()->toString();
}
```

**To reproduce**:

```
$agent = MyAgent::make();
$document = new Document('Lorem Ipsum');
$document->sourceName = 'mysourcename';
$document->sourceType = 'mysourcetype';
$agent->addDocuments([$document]);
```
The class MyAgent is a RAG extension (not implement addDocument but only addDocuments)

This is just my workaround because I only use Qdrant, it’s not a solution.

ilvalerione:
ID is a public property of the Document class, you are free to overwrite it in the process. Anyway the solution should be implemented into the QdrantVectorStore during add documents process, since it is a specific requirements of Qdrant.

ilvalerione:
I'm wondering why this issue is not raised during tests: https://github.com/inspector-apm/neuron-ai/blob/1.x/tests%2FVectorStore%2FQdrantVectorStoreTest.php

nextmwa:
> I'm wondering why this issue is not raised during tests: https://github.com/inspector-apm/neuron-ai/blob/1.x/tests%2FVectorStore%2FQdrantVectorStoreTest.php

I think because in test you set property ID to int example in row 47. 

        $document->id = 1;

ilvalerione:
UUID 4 should be alpha-numeric. How is it compatible with a 64-bit unsigned integer? Do you have a link to Qdrant documentation where they explain this constraint?

nextmwa:
[...]

PR #264, opened 2025-08-14
Fix Qdrant vector store id type #262

nextmwa:
• Removed previous dependency on ramsey/uuid
• Added optional parameter for document ID type in the Document object constructor
• Added private method for generating UUID v4 internally within the Document class

### DIFF
--- a/src/RAG/Document.php
+++ b/src/RAG/Document.php
@@ -4,8 +4,14 @@ declare(strict_types=1);
 
 namespace NeuronAI\RAG;
 
+use const PHP_INT_MAX;
+
 class Document implements \JsonSerializable
 {
+    public const UUID = 'uuid';
+
+    public const INT = 'int';
+
     public string|int $id;
 
     public array $embedding = [];
@@ -18,10 +24,12 @@ class Document implements \JsonSerializable
 
     public array $metadata = [];
 
-    public function __construct(
-        public string $content = '',
-    ) {
-        $this->id = \uniqid();
+    public function __construct(public string $content = '', ?string $idType = null) {
+        $this->id = match ($idType) {
+            'uuid' => $this->generateUuid(),
+            'int' => \random_int(1, PHP_INT_MAX),
+            default => \uniqid(),
+        };
     }
 
     public function getId(): string|int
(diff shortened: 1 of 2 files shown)


ilvalerione:
Don't you think `microtime(true)*10000` is enough to solve the problem?

ilvalerione:
I think keeping only the INT strategy is the best. Documents are created inside data loaders so you typically don't have the chance to pass the id type. Integer should be compatible with all databases.

But extracting a random int does not guarantee uniqueness. Don't you think `microtime(true)*10000` should solve the problem?

```php
public function generateId(): int
{
    return microtime(true)*10000;
}
```

nextmwa:
I don’t think the approach with microtime is correct, because if the document creation is placed inside a loop, it could lead to ID collision issues. However, if you believe this is currently the best solution, it can still be implemented.
But if we don’t require the IDs to be sorted, we can concatenate microtime with a two-digit random integer. This way, we further reduce the likelihood of collisions.

ilvalerione:
I came up with this implementation:

https://github.com/inspector-apm/neuron-ai/blob/fix-doc-unique-id/src/UniqueIdGenerator.php

It handles uniqness with microtime() and add a sequence for fast loops requesting multiple IDs in the same millisecond. 

What do you think?

ilvalerione:
It has been released in 1.17.1

https://github.com/inspector-apm/neuron-ai/releases/tag/1.17.1

nextmwa:
> I came up with this implementation:
> 
> https://github.com/inspector-apm/neuron-ai/blob/fix-doc-unique-id/src/UniqueIdGenerator.php
> 
> It handles uniqness with microtime() and add a sequence for fast loops requesting multiple IDs in the same millisecond.
> 
> What do you think?

Personally, I don’t like using usleep inside a loop because it makes having a powerful machine pointless. However, if for now it’s the only solution without having to drastically change the code, then it’s acceptable.

ilvalerione:
It comes into play only if the maximum sequence is riched, that means the creation of more than 4096 documents in one milliseconds. It's just a consistency strategy, pratically it will never be executed.

PR #271, opened 2025-08-22
Reduce dist size

simPod:
Exclude stuff that are not needed in user's vendor dir for running neuron-ai


ilvalerione:
The "/wiki" directory could be included too.

simPod:
done

PR #287, opened 2025-09-05
Add Provider Support for liteLLM

notdefine:
Add https://www.litellm.ai/ AI Proxy support. liteLLM is API compatible to OpenAI, but needs a different API url

Usage:
```php
new LiteLLM(
  'https://your-littellm.instance.com',
  key: 'sk-...',
  model: 'gpt-4.1',
);
```

### DIFF
--- a/src/Providers/LiteLLM/LiteLLM.php
+++ b/src/Providers/LiteLLM/LiteLLM.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NeuronAI\Providers\LiteLLM;
+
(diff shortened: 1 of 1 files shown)


ilvalerione:
Since it doesn't add any logic it could be integrated with the OpenAILike class: https://github.com/inspector-apm/neuron-ai/blob/2.x/src/Providers/OpenAILike.php

Here is the documentation: https://docs.neuron-ai.dev/components/ai-provider#openailike

notdefine:
I did not see that Class, sorry for the noise

PR #289, opened 2025-09-08
fix: cast document ID to string in addDocuments method

joecharm:
There is an error of the vector ID not being a string when using the Pinecone vector upsert.


ilvalerione:
Fixed in the latest 2.0.7

ilvalerione:
Thank you for the proposal, sorry I closed the PR because I had the change in another branch. Just to avoid conficts.

PR #293, opened 2025-09-11
fix: correct filter key in similarity search request

joecharm:
This fixes filtering vectors when using the Pinecone Vector Store. As per the API docs the request body is 'filter'.

https://docs.pinecone.io/reference/api/2025-04/data-plane/query#body-filter


ilvalerione:
Hi @joecharm Thank you for the fix!

PR #294, opened 2025-09-11
Fix broken links in README.md

AdrienClairembault:
It looks like there are a few broken links in the README, I've updated them to the correct urls.


ilvalerione:
Thank you! Very appreciated.

PR #299, opened 2025-09-17
Add getPageCount method to PdfReader for retrieving PDF page count

joecharm:
A new feature to return the page count when using the PDF Reader. I need it, could be useful for others.

### DIFF
--- a/src/RAG/DataLoader/PdfReader.php
+++ b/src/RAG/DataLoader/PdfReader.php
@@ -127,6 +127,49 @@ class PdfReader implements ReaderInterface
     }
 
     /**
+     * Get the number of pages in the PDF.
+     *
+     * @param string $pdfPath
+     * @return int
+     * @throws DataReaderException
+     */
+    public function getPageCount(string $pdfPath): int
+    {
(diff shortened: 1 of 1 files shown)


ilvalerione:
What's the intended usage? Can you provide a code snippet on how you would use it? Also it should contains at list a unit test here: https://github.com/inspector-apm/neuron-ai/blob/2.x/tests/DataLoader/PdfReaderTest.php

joecharm:
> What's the intended usage? Can you provide a code snippet on how you would use it? Also it should contains at list a unit test here: https://github.com/inspector-apm/neuron-ai/blob/2.x/tests/DataLoader/PdfReaderTest.php

Intended use case is to identify how many pages are being processed by the PdfReader, output can be shown in UI to user or useful for further logic. If you're happy for it to go in then I can add a unit test?

oliverbj:
Weiging in on usefulness. Having the page count can sometimes also be relevant in terms of when and where to perform splitting/chunking.

joecharm:
@ilvalerione  I've now added test_get_page_count() to the PdfReaderTest

PR #303, opened 2025-09-22
Fix Gemini Embeddings config

zliebersbach:
According to the [docs](https://ai.google.dev/api/embeddings#request-body) the `embedding_config` object no longer exists.


ilvalerione:
👍

PR #306, opened 2025-09-24
Fix: Handle MCP tools without input parameters

RauchenwaldC:
## Description
This PR fixes a bug where McpConnector throws an error when working with MCP tools that don't require input parameters.

## Problem
The `createTool()` method assumes all MCP tools have an `inputSchema.properties` field, but according to the MCP specification, tools without input parameters don't need this field. This causes an "Undefined array key 'properties'" error.

## Solution
Added a check for the existence of `inputSchema.properties` before attempting to iterate over it.

## Changes
- Wrapped the properties iteration (lines 115-131) in an `isset()` check
- Added explanatory comment about why this check is necessary

## Testing
- Tested with Laravel Boost MCP server - now works correctly
- Tested with existing MCP tools that have parameters - still works as before
- Both tools with and without input parameters are handled properly

## Affected Systems
This bug prevented integration with:
- Laravel Boost MCP server
- Any other MCP servers that correctly follow the spec

## Example Error (Before Fix)
```
Undefined array key "properties" at vendor/inspector-apm/neuron-ai/src/MCP/McpConnector.php:119
```

This is a non-breaking change that improves compatibility with the MCP specification.

### DIFF
--- a/src/MCP/McpConnector.php
+++ b/src/MCP/McpConnector.php
@@ -112,18 +112,22 @@ class McpConnector
             throw new McpException("Tool response format not supported: {$response['type']}");
         });
 
-        foreach ($item['inputSchema']['properties'] as $name => $prop) {
-            $required = \in_array($name, $item['inputSchema']['required'] ?? []);
+        // Only process properties if they exist in the input schema
+        // Some MCP tools don't require any input parameters
+        if (isset($item['inputSchema']['properties']) && is_array($item['inputSchema']['properties'])) {
+            foreach ($item['inputSchema']['properties'] as $name => $prop) {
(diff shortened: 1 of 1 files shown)


ilvalerione:

ilvalerione:
Released with [2.2.12](https://github.com/inspector-apm/neuron-ai/releases/tag/2.2.12)

You are creating an agent to use the Laravel Boost MCP server, interesting use case. Something like a developer tool?

RauchenwaldC:
@ilvalerione I'm not sure yet if the Laravel-Boost MCP server will make sense in my final project but it was the one I have automatically installed in every Laravel project so I used it for testing and discovered the issue.

Thanks for accepting the PR that quickly and for building Neuron in the first place.

PR #343, opened 2025-10-26
chore: Enhance OpenAIEmbeddingsProvider to handle models without dimensions

giagara:
When using an OpenAI embedding model without dimension, like "text-embedding-ada-002", the api call throw an error.

This PR aims to check if the model needs that field.
Default is yes.


ilvalerione:
A better way could be to make the dimension argument nullable, so developers can configure the provider as per their needs. We don't want to maintain the sync with OpenAI models configuration.

giagara:
@ilvalerione  agree with it but will be a breaking change: how about a `withoutDimensions()` method that can be called?

I've edited the code

giagara:
anyway i get error in github actions:

Run docker/login-action@v3
Error: Username and password required

ilvalerione:
Make the argument nullable with default value at 1024 doesn't break the current API.

For the error on Github actions you need to configure docker credentials as github secrets in your account.

PR #344, opened 2025-10-27
Fix - Ollama httpOptions headers

antioxidanz:
When the httpOptions parameter is filled with headers in the Ollama provider, the following error is thrown:
array_merge(): Argument #1 must be of type array, null given


ilvalerione:
Thank you!

PR #379, opened 2025-11-13
Fix OpenAILike endpoint that do not returns finish_reason

nunomazer:
Fix the error when using an OpenAILike endpoint or model that do not returns "finish_reason" array index


ilvalerione:
missing `finishForToolCall` method

PR #391, opened 2025-11-19
Add Gemini 3 Pro thoughtSignature support

lamberttraccard:
Implements full support for Gemini 3 Pro's mandatory thoughtSignature requirement for function calls.

Changes:
- Add setAnnotation/getAnnotation methods to Tool class for metadata storage
- Extract thoughtSignature from API responses (Gemini.php)
- Include thought_signature when mapping tool calls (MessageMapper.php)
- Preserve thoughtSignature in streaming responses (HandleStream.php)

The thoughtSignature is only included when present in the response from Gemini, maintaining backward compatibility with Gemini 2.5 Pro. Follows official Google documentation for thought signature handling.

### DIFF
--- a/src/Providers/Gemini/Gemini.php
+++ b/src/Providers/Gemini/Gemini.php
@@ -91,9 +91,17 @@ class Gemini implements AIProviderInterface
             }
 
             // Gemini does not use ID. It uses the tool's name as a unique identifier.
-            return $this->findTool($item['functionCall']['name'])
+            $tool = $this->findTool($item['functionCall']['name'])
                 ->setInputs($item['functionCall']['args'])
                 ->setCallId($item['functionCall']['name']);
+
+            // Store thoughtSignature for Gemini 3 Pro compatibility
+            // Gemini 3 Pro requires thoughtSignature to be preserved and sent back
+            if (isset($item['thoughtSignature'])) {
+                $tool->setAnnotation('thought_signature', $item['thoughtSignature']);
+            }
+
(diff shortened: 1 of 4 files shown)


ilvalerione:
Do you have a link to the documentation?

lamberttraccard:
https://ai.google.dev/gemini-api/docs/thought-signatures

lamberttraccard:
I have used my fork in production the last 24h and it works perfectly with all the Gemini 2.5 models and the new Gemini 3 pro preview.

ilvalerione:
With Gemini 2.5 with thinking enabled I didn't found any error with the current implementation. But it becomes strange with Gemini 3. Using `gemini-3-pro-preview` model the streamed response does not contains a finishReason on tool call, so the tools are not executed at all by the agent.

Take a look at the line below, if there is no finishReason in the message the agent drop the tool call execution. Your implementation doesn't solve this issue. How is it working in your environment?
https://github.com/neuron-core/neuron-ai/blob/3.x/src/Providers/Gemini/HandleStream.php#L75

ilvalerione:
Furthermore the solution shouldn't touch the Tool class. It should be implemented on the Message level. The message class has metadata to carry additional information.

ilvalerione:
I released an implementation that supports thuoght signature: [2.8.19](https://github.com/neuron-core/neuron-ai/releases/tag/2.8.19)

Neuron v3 will have full support for reasoning models

PR #414, opened 2025-12-09
fix: FileVectorStore will now create the store dir and file if doesnt exist

amitavroy:
# What does this PR do?

When creating a FileVector store, right now the code will throw an unhandled error `Failed to open stream: No such file or directory`.

As discussed in the discussion link: https://github.com/neuron-core/neuron-ai/discussions/413

It will be a better developer experience if we check for the existance of the directory (recursively) and the file as well.
If they doesn't exist, we try to create them as it's a safe option because there will be no data loss.

And, if the directory create or file create has issues, then we raise an exception.

### DIFF
--- a/src/RAG/VectorStore/FileVectorStore.php
+++ b/src/RAG/VectorStore/FileVectorStore.php
@@ -45,16 +45,6 @@ class FileVectorStore implements VectorStoreInterface
                 );
             }
         }
-
-        // Try to create file if it doesn't exist
-        $filePath = $this->getFilePath();
-        if (!file_exists($filePath)) {
-            if (@file_put_contents($filePath, '') === false) {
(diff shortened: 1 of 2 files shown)


ilvalerione:
I think the file creation is not needed since the file_put_contents will do it automatically. I would keep only the directory creation.

amitavroy:
> file_put_contents

Are you referring to the `appendToFile` method?

Let me know and accordingly I would change the code then.

Thanks for the feedback

ilvalerione:
I'm referring to the like 49 to 56 in the constructor. I would remove this step.

amitavroy:
> I'm referring to the like 49 to 56 in the constructor. I would remove this step.

Done.

PR #426, opened 2025-12-22
fix: gemini structured output REST API compatibility and error handling

BerkayAkpolat:
This PR addresses several critical issues in the `NeuronAI\Providers\Gemini` implementation that prevented `structured()` output from working correctly via the Google Gemini REST API. It also introduces robust error handling to surface API rejections and safety blocks that were previously causing silent crashes.

### 1. REST API Compatibility (`src/Providers/Gemini/HandleStructured.php`)
The Google Gemini REST API has strict requirements for JSON payloads that differed from the library's internal schema generation:
*   **Parameter Naming**: Updated configuration keys from camelCase to **snake_case** (`generation_config`, `response_schema`, `response_mime_type`) to align with the REST API specification.
*   **Schema Transformation**: Overhauled the `adaptSchema` method to recursively transform Neuron's internal schema into a Gemini-compliant format:
    *   **Map Transformation**: Converts the `properties` field from an indexed array of objects into an **associative object map** (required for Google's Protobuf-based JSON parser).
    *   **Type Normalization**: Collapses type arrays (e.g., `["string", "null"]`) into a **single uppercase string** (e.g., `STRING`), as Gemini does not support multi-type arrays.
    *   **Keyword Remapping**: Maps internal PHP types (like `INT`, `BOOL`) to their Gemini equivalents (`INTEGER`, `BOOLEAN`).

### 2. Robust Error Detection (`src/Providers/Gemini/HandleChat.php`)
Previously, the library would crash with an `Undefined array key "candidates"` if the API returned an error or blocked a response due to safety filters.
*   **Explicit Error Checking**: Added logic to catch and throw a `ProviderException` if the API returns an `error` object.
*   **Finish Reason Validation**: The provider now checks the `finishReason` (e.g., `SAFETY`, `OTHER`, `MAX_TOKENS`) and surfaces it in the exception message if a candidate is not returned.

### 3. Stream Integrity (`src/Providers/Gemini/HandleStream.php`)
*   **Mid-Stream Error Handling**: Added a check within the `stream()` loop to detect and throw exceptions if an error occurs mid-generation (e.g., reaching quota limits or triggering safety filters during a response).

## Why this matters
Without these fixes, any attempt to use `structured()` output with the Gemini provider resulted in a `400 Bad Request` from Google due to an invalid JSON payload ("Proto field is not repeating"). These changes ensure that complex Data Transfer Objects (DTOs) with nested arrays and nullable fields can be extracted reliably.

## Verification
*   Tested with complex DTOs containing nested `ArrayOf` mappings.
*   Verified that the `properties` field is correctly serialized as a JSON object `{}` even when empty.
*   Verified that API errors are now properly logged with descriptive messages instead of causing PHP exceptions instead of `ErrorException` crashes.

## Environment
*   **Neuron AI Version**: `v3beta`
*   **PHP Version**: 8.2+
*   **Provider**: Google Gemini REST API (v1beta)

---

### **Bug Report Reference**
The following error is resolved by this PR:
`Gemini API Error: Invalid JSON payload received. Unknown name "type" at 'generation_config.response_schema.properties[0].value.items.properties[1].value': Proto field is not repeating, cannot start list.`

### DIFF
--- a/src/Providers/Gemini/HandleStructured.php
+++ b/src/Providers/Gemini/HandleStructured.php
@@ -38,7 +38,6 @@ trait HandleStructured
                 );
             }
         } else {
-            // If there are no tools, we can enforce the structured output.
             $this->parameters['generationConfig']['responseSchema'] = $this->adaptSchema($response_format);
             $this->parameters['generationConfig']['responseMimeType'] = 'application/json';
         }
@@ -48,21 +47,15 @@ trait HandleStructured
 
     /**
      * Adapt Neuron schema to Gemini requirements.
-     * 1. Remove additionalProperties.
-     * 2. Convert type arrays to single uppercase strings.
-     * 3. Ensure properties is an object map.
      */
     protected function adaptSchema(array $schema): array
     {
-        // 1. Gemini does not support additionalProperties
         if (array_key_exists('additionalProperties', $schema)) {
             unset($schema['additionalProperties']);
         }
 
-        // 2. Fix the 'type' field (Gemini hates type arrays)
         if (array_key_exists('type', $schema)) {
             if (is_array($schema['type'])) {
-                // Collapse ["type", "null"] to "TYPE"
                 foreach ($schema['type'] as $type) {
                     if ($type !== 'null') {
                         $schema['type'] = strtoupper((string) $type);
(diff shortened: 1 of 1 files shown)


ilvalerione:
I merge it to refactor the HandleStructure implementation that is overengeenered

PR #435, opened 2026-01-03
Fixes for Gemini with Structured Output

Leward:
- Fix to support nullable fields
 - Newer Gemini models now support using tool calling and structured output at the same time


Leward:
I created a companion repo to this MR to validate the bug and the fix: https://github.com/Leward/neuron-ai-gemini-bug-repro 

It could be useful to have some integration tests against the actual LLM providers to validate our integration. But that would require setting up a bunch of API keys to run those tests.

PR #442, opened 2026-01-14
Add support for LocalAI reranker

jeff87218:
This PR adds a new `LocalAIRerankerPostProcessor` to support document reranking using a LocalAI rerank API.

####  What’s included
- New `LocalAIRerankerPostProcessor` implementation
- Supports `top_n` and specify reranker models
- PHPUnit tests
- No breaking changes

LocalAI provides an open-source, self-hosted alternative for LLM and reranking services.  
This addition allows users to integrate LocalAI into the RAG pipeline without relying on external APIs.

#### Backward compatibility
- No existing interfaces were modified
- No breaking changes introduced

#### Documentation
New documentation needs to be created

### DIFF
--- a/src/RAG/PostProcessor/LocalAIRerankerPostProcessor.php
+++ b/src/RAG/PostProcessor/LocalAIRerankerPostProcessor.php
@@ -1,0 +1,63 @@
+<?php
+
+namespace NeuronAI\RAG\PostProcessor;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\RequestOptions;
+use NeuronAI\Chat\Messages\Message;
+use NeuronAI\RAG\Document;
+use NeuronAI\RAG\PostProcessor\PostProcessorInterface;
+
+class LocalAIRerankerPostProcessor implements PostProcessorInterface
+{
+    protected Client $client;
+
+    public function __construct(
+        protected string $key,
+        protected string $model = 'cross-encoder',
+        protected int    $topN = 3,
+        protected string $host = 'http://localhost:8080/v1/'
+    )
+    {
+    }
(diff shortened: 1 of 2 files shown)


ilvalerione:
https://docs.neuron-ai.dev/rag/pre-post-processor#localai-reranker

jeff87218:
> https://docs.neuron-ai.dev/rag/pre-post-processor#localai-reranker

Thanks for adding the LocalAI Reranker to the documentation! 🙏  
One small thing: the example is missing the `host` parameter, which is required to configure the LocalAI endpoint.

It would be great to update the example like this:

```php
return [
    new LocalAIPostProcessor(
        key: 'LOCALAI_KEY',
        model: 'LOCALAI_MODEL',
        topN: 3,
        host: 'LOCALAI_URL'
    ),
];
```
This also helps in case LocalAI is not running on localhost and is hosted on a different machine or environment.

ilvalerione:
Finally I changed the host structure moving the "v1" path on the constructor arguments internally. So, the latest version 2.11.1 follow the spec mentioned in the documention: https://docs.neuron-ai.dev/rag/pre-post-processor#localai-reranker

So after updating your dependencies you need to remove "v1" from the constructor if are using a custom host.

jeff87218:
> Finally I changed the host structure moving the "v1" path on the constructor arguments internally. So, the latest version 2.11.1 follow the spec mentioned in the documention: https://docs.neuron-ai.dev/rag/pre-post-processor#localai-reranker
> 
> So after updating your dependencies you need to remove "v1" from the constructor if are using a custom host.

Perfect, thanks for the update!

PR #446, opened 2026-01-20
fix(observability): Add observer lifecycle management to prevent accumulation in long-running processes

SnakeO:
## Summary

This PR fixes an observer accumulation bug in `EventBus` that affects long-running PHP processes such as:

- Laravel queue workers (`php artisan queue:work`)
- PHP-FPM with persistent workers
- Swoole/RoadRunner applications
- PHPUnit test suites

## The Problem

`EventBus` uses static storage for observers. When `observe()` is called, observers are appended to a static array that persists for the lifetime of the PHP process. In short-lived request cycles this is fine, but in long-running processes, observers accumulate:

```php
// Job 1
$agent->observe(new MyObserver($task1));  // 1 observer

// Job 2 (same PHP process)
$agent->observe(new MyObserver($task2));  // Now 2 observers!

// Events fire - BOTH observers receive events
// Observer 1 references stale $task1 - causes bugs
```

### Real-World Impact

We discovered this bug in a Laravel queue worker processing AI automation tasks. After multiple jobs, observers accumulated and caused:

1. **Duplicate log entries** - Each event logged N times (N = job count)
2. **Foreign key violations** - Old observers referenced deleted database records
3. **Memory growth** - Observer array grew unbounded

### Part 1: Observer Lifecycle Methods (Manual Cleanup)

1. **`removeObserver(ObserverInterface $observer): bool`**
   - Explicitly remove a specific observer
   - Returns `true` if removed, `false` if not found
   - Useful for cleanup in `finally` blocks or destructors

2. **`hasObserver(ObserverInterface $observer): bool`**
   - Check if an observer is already registered
   - Prevents duplicate registration

3. **`getObserverCount(): int`**
   - Returns the current observer count
   - Useful for debugging accumulation issues

### Part 2: Scoped Observers (Automatic Cleanup)

**The preferred solution.** Observers registered via `Workflow::observe()` are automatically scoped to that workflow instance and cleaned up when the workflow completes.

```php
// OLD WAY - Manual cleanup required
$workflow = new MyWorkflow();
EventBus::observe($observer);  // Persists forever!
$workflow->init()->run();
EventBus::clear();  // Must remember to call this

// NEW WAY - Automatic cleanup
$workflow = new MyWorkflow();
$workflow->observe($observer);  // Scoped to this workflow
$workflow->init()->run();       // Observer auto-cleared on completion!
```

#### New Scope Methods

4. **`observe(ObserverInterface $observer, ?string $scope = null): void`**
   - Now accepts optional scope parameter
   - Scoped observers can be selectively cleared

5. **`clearScope(string $scope): int`**
   - Clears only observers with the specified scope
   - Returns the number of observers removed
   - Does NOT affect unscoped observers

6. **`getObserverScope(ObserverInterface $observer): ?string`**
   - Returns the scope of an observer, or null if unscoped

### How Scoping Works

```
Workflow A registers Observer 1 (scoped to A)
Workflow B registers Observer 2 (scoped to B)
Global code registers Observer 3 (no scope)

Workflow A completes → Only Observer 1 is removed
Workflow B completes → Only Observer 2 is removed
Observer 3 persists across all workflows
```

### For Queue Workers (Recommended: Scoped Observers)

Simply use `$workflow->observe()` instead of `EventBus::observe()`:

```php
public function handle(): void
{
    $workflow = new MyWorkflow();
    $workflow->observe(new MyObserver());  // Scoped!
    $workflow->init()->run();              // Auto-cleared!
}
```

### For Queue Workers (Alternative: Manual Clear)

If you're not using workflows, call `EventBus::clear()` at the start of each job:

```php
public function handle(): void
{
    EventBus::clear();  // Clear all observers

    $agent = new MyAgent();
    $agent->observe(new MyObserver());
    // ...
}
```

### For Concurrent Workflows

Scoped observers enable safe concurrent execution:

```php
// Both can run concurrently without interfering
$workflowA->observe($observerA);  // Scoped to A
$workflowB->observe($observerB);  // Scoped to B

// When A completes, only $observerA is cleared
// $observerB continues receiving events from B
```

## Backwards Compatibility

This PR is fully backwards compatible:

- `observe($observer)` without scope works exactly as before
- Unscoped observers persist until `removeObserver()` or `clear()` is called
- Existing code that calls `clear()` manually still works
- New scope methods are additive only

## Testing

All 19 tests pass (11 original + 8 new scoped tests):

```
PHPUnit 9.6.31 by Sebastian Bergmann and contributors.

...................                                               19 / 19 (100%)

Time: 00:00.013, Memory: 6.00 MB

OK (19 tests, 49 assertions)
```

### New Test Coverage

- `test_observe_with_scope_tracks_scope`
- `test_observe_without_scope_has_null_scope`
- `test_clear_scope_removes_only_scoped_observers`
- `test_clear_scope_returns_zero_for_unknown_scope`
- `test_remove_observer_also_removes_scope_tracking`
- `test_clear_also_clears_scope_tracking`
- `test_workflow_completion_clears_only_its_observers`
- `test_scoped_observers_receive_events`

## Files Changed

| File | Change |
|------|--------|
| `src/Observability/EventBus.php` | Added scope tracking, `clearScope()`, `getObserverScope()` |
| `src/Workflow/Workflow.php` | `observe()` now scopes to workflow ID |
| `src/Workflow/WorkflowHandler.php` | Auto-calls `clearScope()` after workflow completion |
| `tests/Observability/EventBusTest.php` | Added 8 scoped observer tests |

## Checklist

- [x] Code follows project style guidelines
- [x] Added unit tests for new functionality
- [x] Updated PHPDoc documentation
- [x] No breaking changes
- [x] Tests pass locally

### DIFF
--- a/src/Observability/EventBus.php
+++ b/src/Observability/EventBus.php
@@ -6,6 +6,12 @@ namespace NeuronAI\Observability;
 
 use Inspector\Exceptions\InspectorException;
 
+use function array_search;
+use function array_values;
+use function count;
+use function in_array;
+use function spl_object_id;
+
 /**
  * Global event bus for observability with optional scope support.
  *


ilvalerione:
I don't get the point. In your implementation observers are always registered in the global $observers array and scopes are in a separate array as workflowID or object_id. Then in the clearScope method you only remove observer only by spl_object_id.

Furthermore, the issue is not with multiple jobs running in parallel. Static classes are not shared between processes, but only in the current process. It's more about http requests served with long running processes than background jobs.

Tests don't test the scoped archive in the event bus, that is not what it was implemented by the way, but just observer addition and removal in several scenarios.

If have any additional details, otherwise I'll close this PR and implement this optimization following a different approach.

SnakeO:
Thanks for the feedback! Let me clarify the design decisions:

**On the data structure:**

The dual-array approach (`$observers[]` + `$observerScopes[objectId => scope]`) was chosen for:
1. **O(1) scope lookup** by object ID during removal
2. **Single iteration in `emit()`** - no nested loops, scoping is completely transparent to event emission
3. **Backward compatibility** - unscoped observers work exactly as before

The alternative of grouping as `[scope => observers[]]` would require iterating all scope groups during `emit()`, or flattening on each emit.

But I'm open to alternative approaches if you have a preferred pattern in mind.

**On the problem scope:**

You're correct - the issue is about long-running processes, not separate PHP processes. My PR description mentions PHP-FPM and Swoole specifically. The queue worker example was just one scenario - the core issue is static observer accumulation in any persistent process.

**On tests:**

The tests do cover scoped behavior:
- `test_clear_scope_removes_only_scoped_observers` - verifies selective removal (4 observers, clear scope A, only A's 2 are removed)
- `test_workflow_completion_clears_only_its_observers` - integration test simulating multiple concurrent workflows cleaning up independently

Would you like me to add a test that simulates multiple sequential workflow runs in the same process to demonstrate accumulation prevention more explicitly?

**Alternative approaches:**

If you'd prefer a different implementation, I'm happy to:
1. Restructure to a grouped `[scope => observers[]]` array if you prefer that pattern
2. Follow any existing architectural patterns in the codebase I may have missed
3. Step back and let you implement your preferred approach

What would be most helpful?

ilvalerione:
Since scoping is completely transparent to event emission, it means that when workflowA fires an event it will be propagated to the observers registered for workflowB. Am I wrong?

SnakeO:
You're correct. Scoping is for **lifecycle management**, not **event filtering**. When workflowA fires an event, all registered observers (including workflowB's) will receive it.

The primary problem this PR solves is **sequential accumulation** in long-running processes:

```php
// Job 1: Process email
$workflow->observe(new MyObserver($task1));  // 1 observer
$workflow->run();                            // Observer cleared on completion

// Job 2: Process another email (same PHP process)
$workflow->observe(new MyObserver($task2));  // Still just 1 observer!
```

For **concurrent workflows** in the same process, you're right that cross-pollination occurs. If that's a concern, there are two options:

**Option A: Accept cross-observer events** (current behavior)
Observers can check if an event is relevant to them:
```php
public function onAgentFinish(AgentFinish $event): void
{
    if ($event->agent !== $this->expectedAgent) {
        return;  // Ignore events from other workflows
    }
    // Handle event
}
```

**Option B: Add event filtering** (future enhancement)
`emit()` could accept a scope parameter, and only matching observers receive events:
```php
EventBus::emit($event, scope: $workflowId);  // Only scoped observers receive
```

I left scope-based filtering out intentionally to keep this PR focused on the accumulation bug. Would you prefer I add Option B, or is Option A sufficient for the typical use case?

ilvalerione:
The problem is the case of async execution: https://docs.neuron-ai.dev/neuron-v3/agent/async

In that case I think the static nature of event bus could create the mixing of event handling. In the sync use case it could be enough to clear the bus after the agent ends. Since Neuron supports parallel async running it should be a solution also for this scenario.

ilvalerione:
I finally came up with a solution using scoped observers. It's already available in the latest commit. Thank you for pointing it out, it led to a solution.

PR #459, opened 2026-02-06
Use multipart/form-data for transcription

semhoun:
OpenAI api for transcriptions (https://platform.openai.com/docs/api-reference/audio/createTranscription) use multipart/form-data and not json body. In my tests, json body don't work.

I also fixed a typo error in class name.


ilvalerione:
What issues have you encountered with the current implementation?

semhoun:
I got an error, saying file is not provided. So I look at openai doc, I see they use mulltipart for uploading the file.
It's seem logical, as file body is on json without base64 encoding

ilvalerione:
Are you passing the AudioContent block to the message?
https://docs.neuron-ai.dev/neuron-v3/providers/audio#speech-to-text

Can you paste here a snippet of the code you are using to submit the audio file?

semhoun:
You're last commits, fix the speech-to-text.
Thanks for the fix.

PR #463, opened 2026-02-09
Implementation 462

izdrail:
Implements what was discussed 
https://github.com/neuron-core/neuron-ai/issues/462

### DIFF
--- a/tests/Tools/SearXNGTest.php
+++ b/tests/Tools/SearXNGTest.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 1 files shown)


ilvalerione:
How `$baseUrl` should look like?

izdrail:
> How `$baseUrl` should look like?

It should be the url of the instance searxng 
Information about the api can be found at https://docs.searxng.org/dev/search_api.html

ilvalerione:
If it's only about just one tool, the Toolkit is not needed.

ilvalerione:
For such a simple use case you can use thi implementation in your code, or ship it as a separate composer package.

PR #470, opened 2026-02-12
Fix Gemini 3 function-call detection in HandleChat

### LINKED ISSUE #469: Gemini HandleChat only checks parts[0] for functionCall — breaks Gemini 3 tool calls
## Description

The non-streaming Gemini `HandleChat` trait determines whether a response contains a tool/function call by inspecting only the **first element** of the `parts` array:

```php
if (array_key_exists('functionCall', $parts[0]) && !empty($parts[0]['functionCall'])) {
```

https://github.com/neuron-core/neuron-ai/blob/2.x/src/Providers/Gemini/HandleChat.php#L95

Gemini 3 models (e.g. `gemini-3-pro-preview`, `gemini-3-flash-preview`) can return **text or thought parts before the `functionCall` part**, for example:

```json
{
  "candidates": [{
    "content": {
      "parts": [
        { "text": "Let me call the tool." },
        { "functionCall": { "name": "my_tool", "args": { "input": "value" } }, "thoughtSignature": "sig-abc" }
      ]
    },
    "finishReason": "STOP"
  }]
}
```

When this happens, `$parts[0]` is the text part (`{ "text": "..." }`), the `functionCall` check fails, and the response is treated as a plain `AssistantMessage` — **the tool call is silently ignored**.

See also: https://ai.google.dev/gemini-api/docs/thought-signatures

## Note

The **streaming** path (`HandleStream`) already handles this correctly — `hasToolCalls()` iterates over all parts. Only the non-streaming `HandleChat` path is affected.

Additionally, `Gemini::createToolCallMessage()` uses `array_filter($tools)` without `array_values()`, so when text/thought parts precede `functionCall` parts, the resulting tools array can have non-sequential keys (e.g. `[2 => $tool]` instead of `[0 => $tool]`).

## Expected Behavior

`HandleChat` should scan **all** parts for a `functionCall` key (like `HandleStream::hasToolCalls()` already does), and `createToolCallMessage()` should reindex the filtered tools array with `array_values()`.

manuelkiessling:
Closes #469.

## Problem

The non-streaming `HandleChat` trait only checks `parts[0]` for a `functionCall` key. Gemini 3 models (e.g. `gemini-3-pro-preview`, `gemini-3-flash-preview`) can return text or thought parts *before* the `functionCall` part — see [thought signatures](https://ai.google.dev/gemini-api/docs/thought-signatures). When this happens, the tool call is silently ignored and the response is treated as a plain `AssistantMessage`.

The streaming path (`HandleStream::hasToolCalls()`) already handles this correctly by iterating over all parts.

## Changes

**`HandleChat.php`** — Scan all parts for `functionCall` instead of only `parts[0]`, matching the approach used in `HandleStream::hasToolCalls()`.

**`Gemini.php`** — Wrap `array_filter($tools)` with `array_values()` in `createToolCallMessage()` so the tools array is always 0-indexed. Without this, when text/thought parts precede `functionCall` parts, `array_map` + `array_filter` produces non-sequential keys (e.g. `[1 => $tool]`).

**`GeminiTest.php`** — Five new test cases:
- Function call in first part (regression guard)
- Function call after a text part (the main fix)
- Multiple function calls after a text part
- No function call returns `AssistantMessage`
- `thoughtSignature` metadata is preserved

All existing tests continue to pass.

### DIFF
--- a/src/Providers/Gemini/Gemini.php
+++ b/src/Providers/Gemini/Gemini.php
@@ -18,6 +18,7 @@ use NeuronAI\Tools\ToolInterface;
 
 use function array_filter;
 use function array_map;
+use function array_values;
 
 class Gemini implements AIProviderInterface
 {
@@ -105,9 +106,11 @@ class Gemini implements AIProviderInterface
                 ->setCallId($item['functionCall']['name']);
         }, $message['parts']);
 
+        // array_values() reindexes so tools are always a 0-based sequential array,
+        // even when text/thought parts precede functionCall parts.
         $result = new ToolCallMessage(
             $message['content'] ?? null,
-            array_filter($tools)
+            array_values(array_filter($tools))
         );
 
         if ($signature !== null) {
--- a/src/Providers/Gemini/HandleChat.php
+++ b/src/Providers/Gemini/HandleChat.php
@@ -93,7 +93,18 @@ trait HandleChat
 
                 $parts = $content['parts'];
 
-                if (array_key_exists('functionCall', $parts[0]) && !empty($parts[0]['functionCall'])) {
+                // Scan ALL parts for functionCall, not just parts[0].
+                // Gemini 3 models may return text or thought parts before functionCall parts.
+                // See https://ai.google.dev/gemini-api/docs/thought-signatures
+                $hasFunctionCall = false;
+                foreach ($parts as $part) {
+                    if (array_key_exists('functionCall', $part) && !empty($part['functionCall'])) {
+                        $hasFunctionCall = true;
+                        break;
+                    }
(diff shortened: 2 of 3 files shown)


ilvalerione:
Edited and released as [2.12.4](https://github.com/neuron-core/neuron-ai/releases/tag/2.12.4)

ebrahimpleite:
Please fix this in version 3.x as well

ilvalerione:
It should works in v3

https://github.com/neuron-core/neuron-ai/blob/3.x/src%2FProviders%2FGemini%2FHandleChat.php#L101

PR #473, opened 2026-02-17
Add FactorialTool to CalculatorToolkit

kristoffabjan:
- Added missing FactorialTool to the CalculatorToolkit                                                   
  - The tool was already implemented but not registered in the toolkit's provide() method


ilvalerione:
Ah ok, I didn't get it was already there.

PR #474, opened 2026-02-18
fix(approval-request): Convert actions to iterable entity when restoring from persistent storage

kristoffabjan:
## Summary                                                                                                                             

  - Fixed `src/Workflow/Interrupt/ApprovalRequest.php`  `fromArray()` to validate and convert `actions` to an array when they come back as a JSON string after a `jsonSerialize()`
  round-trip
  - Added a guard for missing `actions` key to avoid iterating over undefined data

  ## Problem

  When `ApprovalRequest` is serialized via `jsonSerialize()` and stored on a persistent medium (e.g. a database), restoring it via
  `fromArray()` fails because `actions` comes back as a JSON string instead of an array, causing:

  foreach() argument must be of type array|object, string given
  at packages/neuron-ai/src/Workflow/Interrupt/ApprovalRequest.php:117

  The original implementation assumed `actions` was always an iterable array, which is not guaranteed after persistence.

  ## Fix

  Validate the `actions` value before iterating — if it's a JSON string, decode it first:

### DIFF
--- a/src/Workflow/Interrupt/ApprovalRequest.php
+++ b/src/Workflow/Interrupt/ApprovalRequest.php
@@ -6,6 +6,8 @@ namespace NeuronAI\Workflow\Interrupt;
 
 use function array_filter;
 use function array_values;
+use function is_array;
+use function json_decode;
 use function json_encode;
 
 class ApprovalRequest extends InterruptRequest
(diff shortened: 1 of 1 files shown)


ilvalerione:
I also added dedicated unit tests for this class. Thank you for the contributions.

PR #475, opened 2026-02-18
feat(workflow-middleware): Add SkipRemainingMiddlewareException to terminate before-middleware chain on demand

kristoffabjan:
## Problem

  The workflow middleware chain for nodes currently runs all registered middleware
  unconditionally. This becomes a UX problem when middleware are ordered to
  enforce a "fail-fast" policy.
  Attaching concrete example and issue I faced in my business problem. Proposed solution solved it.

  **Concrete example** — a `ToolNode` with two middleware:

  1. `ToolAuthorizationMiddleware` — checks Laravel permissions; on failure,
     replaces the tool callable with a `ToolRejectionHandler` (error stub)
  2. `ToolApproval` — prompts the human-in-the-loop for approval before execution

  When authorization fails, middleware No.1 correctly marks the tool as rejected.
  But middleware No.2 still runs and prompts the user for approval — even though the
  tool will return an authorization error regardless of what the user decides.
  **The approval prompt is meaningless and confusing.**

  There is currently no framework-level mechanism for a middleware to say
  "I've done what I need to, skip the rest and proceed to the node."

  ## Solution

  A new signal exception: `NeuronAI\Exceptions\SkipRemainingMiddlewareException`.

  When thrown from a middleware's `before()` method, `Workflow::runBeforeMiddleware()`
  catches it, breaks the loop, and proceeds directly to node execution. `after()`
  methods are **not** affected — they still run for all middleware (cleanup/logging
  semantics are preserved).

  ```php
  // In any WorkflowMiddleware::before() implementation:
  throw new SkipRemainingMiddlewareException('Authorization failed, skipping remaining middleware');
 ```

  Key design decisions

  - Exception carries no payload — it's a pure control-flow signal. Middleware
  is responsible for preparing state (e.g. calling $tool->setCallable(new ToolRejectionHandler(...)))
  before throwing, consistent with the existing ToolApproval::handleRejectedTool() pattern.
  - Multi-tool events: loop all tools first, throw once — middleware must
  iterate all tools, apply rejection handlers to each unauthorized one, then
  throw a single SkipRemainingMiddlewareException after the full loop. This
  ensures no tool escapes the authorization check.
  - after() is unaffected — skip only breaks the before() loop.
  - Fully backward compatible — existing middleware that don't throw this
  exception continue to work exactly as before.

  Changes

  - src/Exceptions/SkipRemainingMiddlewareException.php — new signal exception
  - src/Workflow/Workflow.php — runBeforeMiddleware() catches the exception
  and breaks the loop
  - Tests covering: skip breaks before() chain, after() still runs for all
  middleware, normal chain unaffected when no skip is thrown

  Usage pattern

 ```php
  class ToolAuthorizationMiddleware implements WorkflowMiddleware
  {
      public function before(NodeInterface $node, Event $event, WorkflowState $state): void
      {
          if (!$event instanceof ToolCallEvent) {
              return;
          }

          $anyFailed = false;

          foreach ($event->toolCallMessage->getTools() as $tool) {
              if (!$tool instanceof AuthorizableToolInterface) {
                  continue;
              }

              try {
                  $tool->authorizeToolExecution($tool->getInputs());
              } catch (ToolAuthorizationException $e) {
                  $tool->setCallable(new ToolRejectionHandler($e->getMessage()));
                  $anyFailed = true;
              }
          }

          if ($anyFailed) {
              throw new SkipRemainingMiddlewareException('One or more tools failed authorization');
          }
      }
  }
 ```

  With this in place, a ToolApproval middleware registered after
  ToolAuthorizationMiddleware will never prompt the user for a tool that has
  already been rejected.

### DIFF
--- a/tests/Workflow/SkipRemainingMiddlewareTest.php
+++ b/tests/Workflow/SkipRemainingMiddlewareTest.php
@@ -4,22 +4,24 @@ declare(strict_types=1);
 
 namespace NeuronAI\Tests\Workflow;
 
+use NeuronAI\Exceptions\SkipRemainingMiddlewareException;
 use NeuronAI\Workflow\Events\Event;
 use NeuronAI\Workflow\Events\StartEvent;
 use NeuronAI\Workflow\Events\StopEvent;
-use NeuronAI\Exceptions\SkipRemainingMiddlewareException;
 use NeuronAI\Workflow\Middleware\WorkflowMiddleware;
 use NeuronAI\Workflow\Node;
 use NeuronAI\Workflow\NodeInterface;
 use NeuronAI\Workflow\Workflow;
 use NeuronAI\Workflow\WorkflowState;
 use PHPUnit\Framework\TestCase;
+use stdClass;
 
 class SingleStepNode extends Node
 {
     public function __invoke(StartEvent $event, WorkflowState $state): StopEvent
     {
         $state->set('node_executed', true);
+
         return new StopEvent('done');
     }
 }
@@ -28,11 +30,13 @@ class SkipRemainingMiddlewareTest extends TestCase
 {
     public function testSecondMiddlewareBeforeIsNotCalledWhenFirstThrowsSkip(): void
     {
-        $tracker = new \stdClass();
+        $tracker = new stdClass();
         $tracker->calls = [];
 
-        $firstMiddleware = new class($tracker) implements WorkflowMiddleware {
-            public function __construct(private \stdClass $tracker) {}
+        $firstMiddleware = new class ($tracker) implements WorkflowMiddleware {
+            public function __construct(private stdClass $tracker)
+            {
+            }
 
             public function before(NodeInterface $node, Event $event, WorkflowState $state): void
             {
@@ -46,8 +50,10 @@ class SkipRemainingMiddlewareTest extends TestCase
             }
         };
 
-        $secondMiddleware = new class($tracker) implements WorkflowMiddleware {
-            public function __construct(private \stdClass $tracker) {}
+        $secondMiddleware = new class ($tracker) implements WorkflowMiddleware {
+            public function __construct(private stdClass $tracker)
+            {
+            }
 
             public function before(NodeInterface $node, Event $event, WorkflowState $state): void
             {
@@ -72,11 +78,13 @@ class SkipRemainingMiddlewareTest extends TestCase
 
     public function testAfterMethodRunsForAllMiddlewareEvenAfterSkip(): void
     {
-        $tracker = new \stdClass();
+        $tracker = new stdClass();
         $tracker->calls = [];
 
-        $firstMiddleware = new class($tracker) implements WorkflowMiddleware {
-            public function __construct(private \stdClass $tracker) {}
+        $firstMiddleware = new class ($tracker) implements WorkflowMiddleware {
+            public function __construct(private stdClass $tracker)
+            {
+            }
 
             public function before(NodeInterface $node, Event $event, WorkflowState $state): void
             {
@@ -90,8 +98,10 @@ class SkipRemainingMiddlewareTest extends TestCase
             }
         };
 
-        $secondMiddleware = new class($tracker) implements WorkflowMiddleware {
-            public function __construct(private \stdClass $tracker) {}
+        $secondMiddleware = new class ($tracker) implements WorkflowMiddleware {
+            public function __construct(private stdClass $tracker)
+            {
+            }
 
             public function before(NodeInterface $node, Event $event, WorkflowState $state): void
             {
@@ -130,11 +140,13 @@ class SkipRemainingMiddlewareTest extends TestCase
 
     public function testNormalMiddlewareChainIsUnaffectedWithoutSkip(): void
     {
-        $tracker = new \stdClass();
+        $tracker = new stdClass();
         $tracker->calls = [];
 
-        $firstMiddleware = new class($tracker) implements WorkflowMiddleware {
-            public function __construct(private \stdClass $tracker) {}
+        $firstMiddleware = new class ($tracker) implements WorkflowMiddleware {
+            public function __construct(private stdClass $tracker)
+            {
+            }
 
             public function before(NodeInterface $node, Event $event, WorkflowState $state): void
             {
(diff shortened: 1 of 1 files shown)


ilvalerione:
I think authorization must act when the tools are attached to the agent, not during workflow execution. Authorization should condition if the tool must be available to the agent or not.

```php
protected function tools(): array
{
    $tools = [];

    if (auth()->user()->can(...)) {
        $tools[] = EditArticleTool::make();
    }

    return [
        CalendarToolkit::make(),
        ...$tools,
    ];
}
```

Does it fix your scenario? 

To make it more fluent it can be added an helper method at the tool level to decide if it must be available or not based on a certain condition, like:

```php
protected function tools(): array
{
    return [
        CalendarToolkit::make(),

        EditArticleTool::make()->canRun(
            auth()->user()->can(...)
        ),
    ];
}
```

If the `canRun` method get false, the tool will not be included in the list.
https://github.com/neuron-core/neuron-ai/blob/3.x/src/Agent/HandleTools.php#L74

Maybe I didn't get the point. Feel free to share your thoughts.

kristoffabjan:
Some context. Im building AI  Laravel module, that is heavily dependent on Neuron AI 3.x. Currently, i have Agents defined by users via Agent factory and code-defined tools. 
**My main goal is transparent interaction between user an AI model and strict permission layer.** 

To illustrate on concrete example:
I have Agent model/entity, for which I can store its general prompt, provieder, and attach code defined tools/toolkits(classes extending NeuronAI\Tools\Tool and NeuronAI\Tools\Toolkits\AbstractToolkit). Neuron agents are than instantiated via factory, using settings from my Agent model/instance.
I have permissions layer and settings for tools/toolkits that require approval. 
So one use case, as mentioned in this thread is "CreateArticleTool" usage via agent. In this process, user must have both Article model/entity permissions and AI agent also need Approval in order to proceed with tool executon.

So my solution aims at clear and transparent interaction between user and AI. In my proposed solution, my app deterministically and restrictively handles authorization and approval control via Workflow node middlewares, but on the other hand allow execution/completion of Node with dynamically set Node results. This results in transparent communication between user and AI. Because, looking on above example, if user doesn't have permissions to create Article, middleware will detect this, but inject error message to Tool node, that will still execute, but pass error message/array to AI model. This allows interactive flow, where AI model informs you back, that you dont have permissions, that you are required to approve tool usage etc.

**Examples in your comment basically only exclude tools from Agents toolset, so not providing additional context to user, why this tool was ommited from procees/usage.**

Looking forward to your following comments.

Let me add some screenshots of my example:

App defined Agent:
image


**Feedback about missing permissions. "CreateArticleTool" still executed, but returned predictable error array response to AI. Than AI parsed it human readable format in response.**
image

On assigned permissions, authorization middleware passes, stepping to next middleware in chain(Approval).
image

ilvalerione:
Ok, let me give  you some insights because there a couple of things not documented yet.

Neuron v3 provides the interruption mechanism as a primitive not as a concrete/rigid implementation. All the system is tied to the [InterruptRequest](https://github.com/neuron-core/neuron-ai/blob/3.x/src/Workflow/Interrupt/InterruptRequest.php) primitive, that is an abstract class.

[ApprovalRequest](https://github.com/neuron-core/neuron-ai/blob/3.x/src/Workflow/Interrupt/ApprovalRequest.php#L13) is just a built-in implementation of the interruption feature, to cover the classic use case of tool approval, or action approval in general. But, you are free to create your own interruption experience. You just need to extend the abstract `InterruptionRequest` class into a concrete `ToolAuthorizationInterrutpion` to support your use case.

```php
class ToolAuthorizationInterrutpion extends InterruptionRequest
{
    protected array $tools;

    public function jsonSerialize(): array
    {
        return ['message' => $this->message, 'tools' => $this->tools];
    }

    public function setUnauthorizedTools(array $tools): self
    {
        $this->tools = $tools;
        return $this;
    }
}
```

The `ToolAuthorizationMiddleware` must compile your custom interrutpion request and raises a `WorkflowInterruption` exception with it. The workflow will stop the execution and no other middleware will run. The custom interrutpion request will help you design the right UI experience for your use case. Showing the "unauthorized" message, and something like a "Retry" button.

Once you make your application able to manage the interruption process (persistence, resume) it can be easier to add different UI outputs based on the interruption type your agent decide to forward.

Let me know if this is the way to cover your use case or something mismatch.

kristoffabjan:
Sorry for late response. Wasn't able to reply yesterday.

Thanks for thorough explanation. I will dive a bit deeper in Neuron as well, because I feel like i still lack some knowledge.

I kinda miss some functionality to immediately execute Node with certain result(not only via setCallable()). 
But i guess this is getting more into context of my app, rather to context of Neuron and this PR.

**Please feel free to close this PR.** 


**Just to add some additonal context from my use case(if by any chance give you any useful info for further development):**
I am not completely sure what is the best way to cancel Node execution or Node execution with preset like ToolRejectionHandler. 

In case of having "fail-fast" middleware chain on Workflow nodes, if certain middleware in chain fails, but doesn't require HTL or something, i have issue with execution/failing the Node(and workflow) in controlled manner. For example if Node is ToolNode, AI already returned me assistant message with tool_call, meaning AI is expecting response aka tool result. In this case, i struggle to find nice way to kind of like cancel my workflow with sending error array back to AI. Because for example, if ToolAuthorizationMiddleware fails and WorkflowInterrupt is thrown, even if i for example run setCallable(new ToolRejectionHandler($errorMessage) on ToolNode and resume workflow, next middleware in chain will run on Node resume and stop there, despite the fact that this middleware shouldnt even run.

For now, Im thinking about extending [ToolApproval](https://github.com/neuron-core/neuron-ai/blob/3.x/src/Agent/Middleware/ToolApproval.php) with change below. So skipping middleware if another InterrupRequest is resuming the workflow.

```php
// BEFORE:
if ($node->isResuming() && $node->getResumeRequest() instanceof ApprovalRequest) {

// AFTER:
if ($node->isResuming()) {
    if ($node->getResumeRequest() instanceof ApprovalRequest) {
        $this->processDecisions($node->getResumeRequest(), $event);
    }
    return;
}
```

Sorry for the spam 😄

ilvalerione:
No worry, in fact, thank you for the details, this is what can really help us improve the architecture.

Following your case

> For example if Node is ToolNode, AI already returned me assistant message with tool_call, meaning AI is expecting response aka tool result.

This is not really true. The tool call returned by AI will be added to the chat hisotry at the beginning of the ToolNode: https://github.com/neuron-core/neuron-ai/blob/3.x/src/Agent/Nodes/ToolNode.php#L42

This is on purpose, because allows you to use the middleware `before()` hook to also condition this task. If your middleware fails before the `ToolNode`, AI will never know a tool call was asked, because it will not be in the chat history on the next turn.

I recommend to use Inspector to learn how the Agent works internally.

Feel free to write us for further discussion.

PR #482, opened 2026-02-21
feat: add opt-in Anthropic prompt caching support

hudhaifas:
Add support for Anthropic's prompt caching feature with zero breaking changes. All caching behavior is opt-in and requires explicit enablement.

Changes:
- Add cacheWriteTokens and cacheReadTokens to Usage (default 0)
- Add systemPromptBlocks() method for array-based system prompts with cache_control
- Add withPromptCaching() method to enable tool caching
- Support both new and legacy Anthropic cache token API formats
- Extract cache metrics from streaming responses

Tests:
- 7 new tests covering all caching scenarios
- Verify opt-in behavior (no caching by default)
- Verify backward compatibility with string systemPrompt()
- All existing tests pass (894 tests, 3579 assertions)

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ Learn more about Monitoring in the [documentation](https://docs.neuron-ai.dev/ad
 With Neuron, you can switch between [LLM providers](https://docs.neuron-ai.dev/components/ai-provider) with just one line of code, without any impact on your agent implementation.
 Supported providers:
 
-- [Anthropic](https://docs.neuron-ai.dev/providers/ai-provider#anthropic)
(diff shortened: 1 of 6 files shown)


ilvalerione:
I think the cache metrics shouldn't be reported in the `Usage` object because they are specific of the Anthropic API format. You can remove them from the Usage object that is standard for all providers, and attach the cache metrics in the assistant message metadata.  

```php
$message->addMetadata('cacheWriteTokens', $cacheWrite)
    ->addMetadata('cacheReadTokens', $cacheRead);
```

Issue #486, opened 2026-02-23
Failed assertions leak across dataset items causing false failures


Leward:
**Describe the bug**
`EvaluatorRunner` does not reset assertion state between dataset items during evaluation runs. Failed assertions from one dataset item leak into subsequent items, causing correct assertions to be incorrectly reported as failed and inflating the total assertion count.

**To Reproduce**

1. Create an evaluator with multiple dataset items where the first item fails an assertion and subsequent items should pass
2. Run the evaluation: `php vendor/bin/neuron evaluation --path=evaluators`
3. Observe that passing assertions in later items are incorrectly reported as failed

**Minimal reproduction repository:** https://github.com/Leward/neuronai-assertion-state-leak

**Expected behavior**

Each dataset item should have isolated assertion state:
- Test 0: `Assertions: 0 passed, 1 failed`
- Test 1: `Assertions: 1 passed, 0 failed`
- Summary: `Assertions: 2, Passed: 1, Failed: 1`

**Actual behavior**

Assertion state accumulates across dataset items:
```
There were 2 failure(s):

1) Test #0
   Evaluation failed
   Assertions: 0 passed, 1 failed
   Execution Time: 0s

2) Test #1
   Evaluation failed
   Assertions: 1 passed, 1 failed  ❌ Should be: 1 passed, 0 failed
   Execution Time: 0s

Assertion Failure Summary:
--------------------------------------------------
LeakAssertionStateEvaluator:30 - 2 failures in StringContains

FAILURES!
Tests: 2, Passed: 0, Failed: 2, Success Rate: 0%
Assertions: 3, Passed: 1, Failed: 2, Success Rate: 33.3%
            ↑ Should be 2 total assertions, not 3
```

**The Neuron AI version you are using:**
- Version 3.0.0 (also affects v2.12.x)

**Additional context**
`EvaluatorRunner` reuses a single `BaseEvaluator` instance for all dataset items. The evaluator stores assertion state in an internal `RuleExecutor`, which is never reset between items.

PR #487, opened 2026-02-23
Fix assertion state leak across dataset items in evaluations

Leward:
Resolves: #486

Failed assertions from one dataset item were leaking into subsequent items during evaluation runs. The `EvaluatorRunner` reused a single `BaseEvaluator` instance without resetting its internal `RuleExecutor` state, causing assertion counts to accumulate incorrectly.

Changes:
- Add `resetAssertionState()` method to `BaseEvaluator` to clear assertion counters and failures
- Call `resetAssertionState()` in `EvaluatorRunner` before processing each dataset item to ensure isolated assertion state


ilvalerione:

ilvalerione:
It's now available in the latest commit.

PR #489, opened 2026-02-23
Add OpenSearch Vector Store

jamosaur:
Adds OpenSearch as a vector store provider.

Added some basic tests that mirror the elasticsearch tests too.


ilvalerione:
Thank you for this implementation!

PR #501, opened 2026-02-26
Refactor to create collection in a separate method

bytestream:
Allows you to override `createCollection` if you need to initialise indexes etc.


bytestream:
I see a few of the other VectorStores have `createIndex` logic, maybe it just needs adding here too. In my experience, it's needed on `sourceType` and `sourceName` as type `keyword`:

```php
$this->httpClient->request(
  HttpRequest::put(
    uri: '/index',
    body: [
      ['field_name' => 'sourceType', 'field_schema' => ['type' => 'keyword']]
    ]
  )
);
```

PR #503, opened 2026-03-02
Ability to delete documents by sourceType

bytestream:
Allows you to delete documents en masse  by `sourceType`


ilvalerione:
I think we can achieve the same result just making the current delete method a bit smarter with a nullable "sourceName". 

`deleteBySource()` is probably a wrong name for this method that can become just `deleteBy`, accepting source type and a nullable source name as arguments. 

Since `deleteBySource` can't be removed to avoid breaking changes, it can be marked as deprecated, and become just a proxy for the new `deleteBy`.

PR #513, opened 2026-03-17
Allow proper union types inside of structured outputs

jamosaur:
Allows union types to be used with structured output.

Deserializing also checks the data type and tries to cast it to the correct type if it is defined on the structured output class. Behaviour falls back to what was originally there (attempting to cast each type in order).

### DIFF
--- a/src/StructuredOutput/Deserializer/Deserializer.php
+++ b/src/StructuredOutput/Deserializer/Deserializer.php
@@ -5,11 +5,11 @@ declare(strict_types=1);
 namespace NeuronAI\StructuredOutput\Deserializer;
 
 use BackedEnum;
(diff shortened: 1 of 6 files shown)


ilvalerione:
Can you provide code examples of the feature it's intended to add?

ilvalerione:
@jamosaur I would like to better understand this PR. Can you provide more context and examples?

jamosaur:
@ilvalerione my use case for this was having something coming back from the agent that could have been float, int or null depending on some factors.

I just added support for everything rather than only my use case here. Happy for it to not be merged if it's not the direction you'd like to go of course 😄

ilvalerione:
Can you provide a code example about before and after this PR?

PR #534, opened 2026-04-08
Handle reasoning end in handleText method

saithink:
Add reasoning end handling in text chunk processing.

### DIFF
--- a/src/Chat/Messages/Stream/Adapters/AGUIAdapter.php
+++ b/src/Chat/Messages/Stream/Adapters/AGUIAdapter.php
@@ -59,6 +59,15 @@ class AGUIAdapter extends SSEAdapter
 
     protected function handleText(TextChunk $chunk): iterable
     {
+        // Close reasoning if it was started (transition from reasoning to text)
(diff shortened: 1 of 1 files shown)


ilvalerione:
Shouldn't it also include messageId? 

https://docs.ag-ui.com/concepts/events#reasoningend

This proposal make me think that we should also manage the opposite: end the text stream when reasoning start, or handle the end of both, text and reasoning, on tool call. Maybe these END events can be encapsulated in dedicated methods so thay can be called: 

- endReasoning entering in the text stream
- endText entering in the reasoning stream
- both entering in the tool call stream

What do you think?

saithink:
Thanks for your feedback! I totally agree. Adding messageId makes a lot of sense, and encapsulating the END events into dedicated methods will definitely make the state transitions (reasoning/text/tool call) much cleaner. I'll update the PR with these changes shortly.

saithink:
I've updated the code as suggested, including the messageId and the refactored event handling methods. Please take another look!

PR #540, opened 2026-04-10
Handle OpenAI API errors in processChatResult

viicslen:
This pull request adds improved error handling to the `processChatResult` method in `src/Providers/OpenAI/Responses/HandleChat.php`. The main change is to ensure that errors returned from the OpenAI API are detected and reported clearly, making debugging and maintenance easier.

**Error handling improvements:**

* Added a check for an `'error'` key in the `$result` array, throwing a `ProviderException` with the error message if present.
* Added a check to ensure the `'output'` key exists in the `$result` array, throwing a `ProviderException` if it is missing.

### DIFF
--- a/src/Providers/OpenAI/Responses/HandleChat.php
+++ b/src/Providers/OpenAI/Responses/HandleChat.php
@@ -57,11 +57,11 @@ trait HandleChat
      */
     protected function processChatResult(array $result): AssistantMessage
     {
-        if (array_key_exists('error', $result)) {
+        if (isset($result['error']) {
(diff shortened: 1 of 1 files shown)


ilvalerione:
the "empty output" is not necessary on error. I don't think it should raise an exception. It could depends by the stop reason for example.

viicslen:
@ilvalerione As I’m not too familiar with the response structure, I have updated the checks to include only what is necessary for the existing `array_filter($result['output'], ...)` to work. This prevents a **TypeError** from being thrown when the first parameter is `null` instead of an `array`, which is currently happening in some instances.

Alternatively, would it be better to use the null coalescing operator (`$result['output'] ?? []`) instead?

PR #542, opened 2026-04-13
fix: OpenAIImage stream TypeError + error / truncation handling

### LINKED ISSUE #541: OpenAIImage::stream() throws TypeError on first SSE event
**Describe the bug**

Streaming image generation errors out immediately on the first SSE event:

```
TypeError: NeuronAI\Chat\Messages\Stream\Chunks\ImageChunk::__construct(): Argument #1 ($messageId) must be of type string, int given, called in .../src/Providers/OpenAI/Image/OpenAIImage.php on line 157
```

`$line['partial_image_index']` is an int in the OpenAI event payload, but `ImageChunk` takes `string $messageId`, so it throws on the first partial event.

I noticed two more problems in the same loop. `image_generation.completed` isn't handled, so an early stream close can return the last partial as if it were the final image. And `{"type":"error", ...}` events get swallowed.

**To Reproduce**

```php
$provider = new OpenAIImage(
    key: getenv('OPENAI_API_KEY'),
    model: 'gpt-image-1',
    output_format: 'webp',
    parameters: ['partial_images' => 2, 'size' => '1024x1024'],
);

foreach ($provider->stream(new UserMessage('a cat')) as $chunk) {
    // never reaches here — TypeError thrown on first event
}
```

**Expected behavior**

Partial events yield `ImageChunk` instances; the completed event populates the final `AssistantMessage` with usage. Error events surface as `ProviderException`.

**The Neuron AI version you are using:**
 - 3.3.5. Looks like this has been there since the method landed in February.

**Additional context**

OpenAI Images streaming reference: https://developers.openai.com/api/reference/resources/images/generation-streaming-events

PR incoming.

CodeWrap:
## Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)

### What does this PR do?

Fixes three bugs in `OpenAIImage::stream()`.

Main one: it passes `partial_image_index` into `ImageChunk` as the message id. That value is an int, so under `strict_types` it dies on the first SSE event.

Also, `image_generation.completed` wasn't handled, so an early stream close could return the last partial as the final image. And `{"type":"error", ...}` events were ignored instead of throwing.

Public API stays the same. The generator still yields `ImageChunk` for partial events and returns an `AssistantMessage` at the end, but the final image and usage now come from `image_generation.completed`.

Stream method is currently unusable — errors out on the first SSE event from the real API.

### How does this change should be used and documented?

No usage changes. Same call site as before.

## Related Issues

Fixes #541

## Testing

- [x] All existing tests pass
- [x] I have tested this with different PHP versions (if applicable)

Added 7 tests in `tests/Providers/OpenAIImageTest.php` covering partial chunks, completed-only streaming, request payload shape, API error events, truncated streams, missing `b64_json`, and the existing non-streaming `chat()` path.

## Breaking Changes

- [x] No breaking changes

`ImageChunk` untouched. `messageId` is generated once per stream via `UniqueIdGenerator::generateId('msg_')`, matching `OpenAITextToSpeech` / `OpenAISpeechToText`.

## Documentation

- [x] No documentation changes needed

## Checklist

## Additional Notes

OpenAI reference: https://developers.openai.com/api/reference/resources/images/generation-streaming-events

If you'd rather keep this to the type fix, I can split the error/truncation changes into a follow-up PR. LMK.

### DIFF
--- a/src/Providers/OpenAI/Image/OpenAIImage.php
+++ b/src/Providers/OpenAI/Image/OpenAIImage.php
@@ -22,8 +22,10 @@ use NeuronAI\Providers\AIProviderInterface;
 use NeuronAI\Providers\MessageMapperInterface;
 use NeuronAI\Providers\SSEParser;
 use NeuronAI\Providers\ToolMapperInterface;
+use NeuronAI\UniqueIdGenerator;
 
 use function end;
+use function is_string;
 
 class OpenAIImage implements AIProviderInterface
 {
@@ -142,7 +144,8 @@ class OpenAIImage implements AIProviderInterface
             )
         );
 
-        $content = '';
+        $messageId = UniqueIdGenerator::generateId('msg_');
+        $content = null;
         $usage = new Usage(0, 0);
 
         while (! $stream->eof()) {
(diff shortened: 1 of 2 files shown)


ilvalerione:
Why creating a custom msg_id instead of using the provider `$line['partial_image_index']`?

CodeWrap:
Mainly because of the type — `partial_image_index` is an int and `ImageChunk` takes a string, which is the actual TypeError. It also resets per stream so it's not really a stable id. I followed `OpenAITextToSpeech` which already does `UniqueIdGenerator::generateId('msg_')`. Not married to that approach if you have something else in mind.

ilvalerione:
cast to string could be applied. Does the `UniqueIdGenerator::generateId('msg_')` reset per stream too?

CodeWrap:
I think the cast fixes the `TypeError`, but it changes what `messageId` means. `UniqueIdGenerator::generateId('msg_')` is called once per `stream()` call, so yes, a new stream gets a new id. The important part is that within one stream every `ImageChunk` shares that same id. `partial_image_index` is the opposite: it changes for each partial (`"0"`, `"1"`, `"2"`) and then starts over on the next stream. For consumers grouping chunks by `messageId`, that's the difference between one image tile updating as partials arrive and three separate tiles.

PR #543, opened 2026-04-13
Improve serializability of interrupts in workflows

tryvin:
This PR introduces some improvements to the serializability of interrupts.

- When MCP tools were being used, triggering an interrupt from a middleware (like ApprovalRequest) would fail to persist, as serialization would fail due to the socket/curl being open and the presence of a Closure as the `callback` parameter.
- Some databases have an issue with "binary-like" strings (named CockroachDB, even though it's a PSQL "compatible" database), so, instead of saving it as a serialized string, it uses base64 to encode and decode it (with a fallback for keeping it retro-compatible)
 - Use `callable` instead of `Closure` at Toolkit `with` parameters, as Closure can't be serializable and a callable can be "anything" from a PHP perspective, achieving the same result.

I also ran `composer format` across the whole codebase so that PHP-CS-Fixer would pass.


tryvin:
@ilvalerione Quick question: how do you feel about me adding a check to see if the `SerializableClosure` from Laravel is available and wrapping `Closure` with it? At the `with` methods from a Toolkit?

So it would be something like:

```php
public function with(string $class, callable $callback): ToolkitInterface
    {
        if ( class_exists('Laravel\SerializableClosure\SerializableClosure') && $callback instanceof Closure ) {
            $this->with[$class] = new \Laravel\SerializableClosure\SerializableClosure($callback);
        } else {
            $this->with[$class] =  $callback;
        }

        return $this;
    }
```

ilvalerione:
Nope

ilvalerione:
Thank you! Great contribution! I'm going to implement the base_64 bridge also in the generic DatabasePersistence component before releasing it.

ilvalerione:
Refined and released in the [3.3.8](https://github.com/neuron-core/neuron-ai/releases/tag/3.3.8)

tryvin:
Thanks for the improvements @ilvalerione!

PR #566, opened 2026-05-08
Parameter-Aware Tool Call Tracking

tryvin:
This PR introduces custom run key tracking, addressing a limitation where tools could only be tracked by name, not by their invocation parameters.

## Problem

Currently, Neuron tracks tool runs only by tool name. This prevents blocking infinite loops in which an LLM repeatedly calls the same tool with different parameters (e.g., reading a file in chunks with different offsets). The workaround has been to return `PHP_INT_MAX` from `getMaxRuns()` to disable limiting entirely.

## Solution

- Introduced `RunKeyInterface` that tools can opt-in to implement, replacing the previous `ParameterizedRunKeyTool`
- Simplified `getRunKey()` to take no arguments — tools have full control over their key strategy
- Consolidated `AgentState` tracking: `incrementToolRun()`, `getToolRuns()`, `resetToolRuns()` now accept a generic key string instead of requiring a tool name
- Removed separate `*ByKey` methods — a single unified path for all tracking
- Created `TrackByInputs` trait providing input-based key generation (`name:json_encode(inputs)`) as a ready-to-use strategy
- Modified `ToolNode` and `ParallelToolNode` to check for `RunKeyInterface` and use custom keys when available
- Maintained full backwards compatibility — tools without the interface use the tool name as the key

## Implementation

Tools implementing `RunKeyInterface` provide a unique key by returning any string from `getRunKey()`:

```php
public function getRunKey(): string
{
    return $this->getName() . ':' . $this->getInput('file_path') . ':' . $this->getInput('offset');
}
```

Or use the `TrackByInputs` trait for an automatic input-based key:

```php
class ReadFileTool extends Tool implements RunKeyInterface
{
    use TrackByInputs;
}
```

This allows the same tool to be called multiple times with different parameters while still blocking infinite loops with identical keys.

## Testing

- 40 Agent tests pass (84 assertions)
- Full PHPStan level 5 compliance
- Test fixture `TestParameterizedTool` uses the `TrackByInputs` trait, covering the new utility


ilvalerione:
I would take in consideration a different approach.

Instead of adding three new methods on the `AgentState`, we could use the already existing `incrementToolRun()`, `getToolRuns()`, and `resetToolRuns()`, but instead of getting "$toolName", we can use a generic string "$key".

`ParametrizedRunKeyTool` interface becomes `RunKeyInterface`, and the getRunKey() method just return a string without getting input arguments. 

Now Tools can decide to implement this interface to be tracked with a custom key.

If the Tool implements the `RunKeyInterface`, ToolNode and ParallelToolNode use `$tool->getRunKey()`, otherwhise just the tool name. 

This is more generic, allow developers to totally customize their strategy, and it will help me to merge the `getRunKey()` method in the ToolInterface in v4.

You can eventually implement a trait TrackByInputs that implement the getRunKey() method to provide the strategy subject of this PR.

What do you think?

tryvin:
@ilvalerione sounds good, I will implement it!

src/Agent/AgentState.php
```diff
@@ -32,13 +32,21 @@ public function setChatHistory(ChatHistoryInterface $chatHistory): AgentState
         return $this;
     }
 
+    /**
+     * @param string $toolName The tool name for regular tools, or a custom run key
+     *                         when the tool implements RunKeyInterface.
+     */
     public function incrementToolRun(string $toolName): void
```
tryvin, line 39: @ilvalerione I didn't rename the `toolName` param so it wouldn't break anyone calling these methods with named params

ilvalerione: I will change it in the next major,

tryvin: @ilvalerione Is there a roadmap for v4 available?

ilvalerione: too early now. I accumulated a bunch of stuff, so I'm starting to implement these changes before working on the new things I have in mind. A lot of incredible new features though.

ilvalerione:
I also made a video on this PR this morning :)

https://www.youtube.com/watch?v=LhoOQD2Jlc8

Feel free to share.

tryvin:
Mom, I'm famous 😂 

But to answer, yeah, the idea is basically as the agentic workflow I'm working on tries to use the reinforcement learning that the frontier models have on Agentic Coding, instead of having thousands of tools for each use case, I expose a "Virtual File system" where they can read and write files as they please, and as expected, in a long agentic workflow the model might call the read and write tools a lot of times, even though they are reading or writting to different files/parts of the files.

The same thing happens for web search and web fetching, as these usually require multiple web searches and retrieving web content in chunks, so I don't explode the context window; again, the same tools are being called over and over.

Usually, you wouldn't see this happen with "smaller tasks", where the agent does a couple of calls and completes the task

ilvalerione:
Yes, we also want to design for making scalable use cases possible with easy.

PR #571, opened 2026-05-13
chore: support for interrupting the workflow from a tool call, resumi…

tryvin:
# Tool-Driven Workflow Interrupts

This PR introduces tool-driven workflow interrupts, enabling tools to pause the workflow and request external input (e.g., human approval) on their own initiative — not just through middleware.

## Problem

Previously, `WorkflowInterrupt` could only be triggered from middleware (`ToolApproval`) or manually within a node's `__invoke`. Tools were pure computation units with no mechanism to signal that they needed external input during execution. This made patterns like human-in-the-loop approval impossible at the tool level.

## Solution

- **`HasInterrupt`** — opt-in interface (like `HasRunKey`) that tools implement to signal interrupts. Extends `ToolInterface` so tools implementing it always have `getName()`, etc.
- **`InterruptHandler`** — trait providing property storage and getter/setter for interrupt and resume requests
- **`ToolsInterruptRequest`** — single envelope wrapping one or more tool interrupt requests, used by both sequential and parallel execution
- **`ToolNode`** — detects interrupt request after `$tool->execute()` and throws `WorkflowInterrupt`; injects resume request into tools on workflow resume, skipping chat history to avoid duplication; clears serialized old interrupt requests on resume
- **`ParallelToolNode`** — collects interrupt requests from all forked tools and merges them into a single `ToolsInterruptRequest` before throwing

### Multi-Step State Tracking

Tool properties survive serialization across interrupt/resume cycles, enabling tools to track progress through multiple confirmation steps using their own properties:

```php
class MultiStepTool extends Tool implements HasInterrupt
{
    use InterruptHandler;

    private int $step = 0;

    public function __invoke(mixed ...$params): string
    {
        if ($this->getResumeRequest() !== null) {
            $this->step++;
        }

        if ($this->step >= count($this->steps)) {
            return 'All steps completed';
        }

        $this->setInterruptRequest(new ApprovalRequest($this->steps[$this->step]));
        return '';
    }
}
```

### Implementation

Tools implement `HasInterrupt` and use the `InterruptHandler` trait:

```php
class PurchaseTool extends Tool implements HasInterrupt
{
    use InterruptHandler;

    public function __invoke(float $amount): string
    {
        if ($this->getResumeRequest() !== null) {
            return 'Approved: $' . $amount;
        }

        $this->setInterruptRequest(new ApprovalRequest(
            "Approve purchase of \${$amount}?",
            [new Action('approve', 'Approve', true)]
        ));

        return '';
    }
}
```

> **Note on "headless tools":** The library intentionally ships no built-in frontend-delegating tool. That is a well-known pattern ("headless tool") that belongs in user applications and may require integration with streaming adapters / frontend protocols. Users can build their own using `HasInterrupt` + `InterruptHandler`.

## Testing

15 new tests covering:
- `ToolsInterruptRequest` (add/get/count)
- `HasInterrupt` signaling (ToolNode)
- Resume injection
- Regular tool passthrough
- Multi-step state tracking (direct invoke, ToolNode, serialize/unserialize round-trip)

Test fixture `MultiStepTestTool` verifies state survives serialization across interrupt/resume cycles.

Full PHPStan level 5 compliance.


ilvalerione:
So the intention is to be able to ask an interruption from inside the tools. FrontendTool is just a built-in use case. Right?

tryvin:
@ilvalerione Yeah, the idea is that a tool would be able to interrupt the workflow as it pleases. In this case, it was the FrontendTool, but it can be something like "ConfirmPlan", or "RunCode", and then the tool could do some sort of analysis, and if the code is dangerous, interrupt and ask the user what it thinks.

That way, you keep the interruption at the same place the code actually lives, and don't need to rely on a middleware to figure out things that might never be called.

Another use case is a "multi-step confirmation", for example, I have a GenerateImage tool, where the user sends a prompt, the assistant calls the tool, the tool sends this prompt to another LLM to enhance it, and then, as it was before, I would return the prompt to the assistant and ask it to confirm with the user and call the same tool again with the confirmedPrompt. So with the tool being able to interrupt, I can receive a single prompt, enhance, interrupt, user confirms or denies, or customize, and then I proceed with the image generation.

ilvalerione:
Ok, it sounds interesting. At a first look I think there are other ways to explore this feature based on the underlying workflow architecture. I'll give you a feedback.

tryvin:
@ilvalerione Sure, lemme know what you think, and we can update it

tryvin:
@ilvalerione Any thoughts on this? No pressure, want to know what the direction would be so I can update my fork

ilvalerione:
Hey @tryvin, I haven't had the time to sit down properly with the proposal yet, and I don't want to give you a direction off the top of my head and then change my mind once I look at it carefully. I'll come back with concrete feedback so you can move forward. Thanks for the patience.

tryvin:
@ilvalerione Lemme know when you can take a look here :)

tryvin:
@ilvalerione :'(

ilvalerione:
I didn't forget :) I was exploring different impacts and technical approaches. I have a first feedback about this PR: Remove the "Frontend Tool" feature from this proposal.

What you call "frontent tool" is a well known pattern named "headless tool" that allows backend agents to control frontend applications, and it also requires to be integrated with [streaming adapters](https://docs.neuron-ai.dev/agent/streaming#stream-adapters) to make it work with frontend protocols.

tryvin:
@ilvalerione Sorry for taking so long, it's updated!

ceskyDJ:
Oh, I missed this, thanks for your effort! This is something I need, too. Currently, I use middleware for that, and it's kinda hacky, this would be a much cleaner alternative!

tryvin:
@ilvalerione Any news?

ilvalerione:
There are several incosistencies in this implementation, but I would like to support this feature so I opened a branch for it: https://github.com/neuron-core/neuron-ai/tree/tool-interrupt

I'm going to test it in different ways and eventually it will be integrated into the main repo.

PR #573, opened 2026-05-14
add Alibaba & Deepseek ReasoningContent

tw2066:


ilvalerione:
You already have `getTextBlocks()` to get text information. If you want to extract the string directly you can just use `getContent()` that now returns only the text part with no reasing: https://github.com/neuron-core/neuron-ai/blob/3.x/src/Chat/Messages/Message.php#L108

tw2066:
Modified

ilvalerione:
Why using index "-1"?

tw2066:
Both reasoningContent and Content equal 0, use -1 for differentiation.

PR #574, opened 2026-05-14
fix(gemini): add toolConfig AUTO mode to prevent built-in tool invocation

marceloandrader:
## Problem

When Gemini thinking models (e.g. 2.5 Pro) are given function tools, they can spontaneously invoke built-in provider-side tools like `run` (code execution). Since these are never registered in NeuronAI's tool list, `findTool()` throws a `ProviderException`:

```
The model is asking for a non-existing tool: run. You could try writing more verbose tool descriptions and prompts to help the model in the task.
```

## Fix

Set `toolConfig.functionCallingConfig.mode = AUTO` in the Gemini chat request whenever function tools are present. This explicitly tells Gemini to use standard function-calling behaviour without activating built-in provider tools like code execution or Google Search.

`AUTO` is Gemini's documented default for function calling and does not restrict which of our declared functions the model may call — it simply prevents it from reaching for built-in tools that NeuronAI has no knowledge of.

## References

- [Gemini function calling modes](https://ai.google.dev/gemini-api/docs/function-calling#function-calling-modes)

### DIFF
--- a/src/Providers/Gemini/HandleChat.php
+++ b/src/Providers/Gemini/HandleChat.php
@@ -14,12 +14,10 @@ use NeuronAI\Chat\Messages\Usage;
 use NeuronAI\Exceptions\HttpException;
 use NeuronAI\Exceptions\ProviderException;
 use NeuronAI\HttpClient\HttpRequest;
-
 use NeuronAI\Tools\ToolInterface;
 
 use function array_filter;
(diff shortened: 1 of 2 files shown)


ilvalerione:
It's more efficient with a foreach I think. Once you find the first ToolInterface set the body and break. It will be one iteration most of the times.

PR #575, opened 2026-05-18
Optimize the disconnect method in StdioTransport

tw2066:
Added use declaration for TypeError
Wrapped process termination logic in a try-catch block Caught TypeError exceptions to prevent process cleanup failure Reset the pipes property to null after exception handling

### DIFF
--- a/src/MCP/StdioTransport.php
+++ b/src/MCP/StdioTransport.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace NeuronAI\MCP;
 
-use TypeError;
 use function array_merge;
 use function escapeshellarg;
 use function fclose;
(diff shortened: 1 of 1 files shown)


ilvalerione:
More context is needed. Please clearly explain the problem you met, how to reproduce the scenario, why the current implementation doesn't represent real cases, and why you fixed this way.

tw2066:
Currently, in the hyperf framework, a type error will be reported when running commands

tw2066:
Optimize the disconnect method in StdioTransport. use __destruct

ilvalerione:
`__destruct()` delegates to `disconnect()` as a safety net.

Something like:

```php
public function disconnect(): void
{
  if (!is_resource($this->process)) {
	  return;
  }

  foreach ($this->pipes as $pipe) {
	  if (is_resource($pipe)) {
		  fclose($pipe);
	  }
  }

  $status = proc_get_status($this->process);
  if ($status['running'] && function_exists('proc_terminate')) {
	  proc_terminate($this->process);
          usleep(500000);
  }

  proc_close($this->process);
  $this->process = null;
  $this->pipes = null;
}

public function __destruct()
{
  $this->disconnect();
}
```

tw2066:
Modified

PR #587, opened 2026-06-11
Fix property validation incorrectly rejects null on nullable properties

bytestream:
I have a class with:
```php
    #[SchemaProperty(
        description: 'The category that best describes the reason for escalation. Choose from the predefined categories based on the conversation context.',
    )]
    #[Enum(class: EscalationCategory::class)]
    public ?EscalationCategory $category = null;
```

It's failing validation because the LLM returned `null`:
```
structured-validated {
  "class":"...",
  "json":"{\"handoff\":false,\"reason\":\"...\",\"category\":null,\"handoff_message_unauthenticated\":\"\",\"handoff_message\":\"\"}",
  "violations": [
    "category must be one of the following allowed values: a, b, c."
  ]
}
```

This PR makes changes so that `null` is permitted on `nullable` properties. Validation rules will be skipped

### DIFF
--- a/src/StructuredOutput/Validation/Rules/Enum.php
+++ b/src/StructuredOutput/Validation/Rules/Enum.php
@@ -22,12 +22,14 @@ class Enum extends AbstractValidationRule
     /**
      * @param array|null $values List of allowed values. Cannot be used with `enum`.
      * @param string|null $class Enum class name to extract choices from. Cannot be used with `choices`.
+     * @param bool $nullable Whether null is a valid value.
      *
      * @throws StructuredOutputException if both `choices` and `enum` are provided or if `enum` is not valid.
      */
     public function __construct(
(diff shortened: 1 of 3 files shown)


ilvalerione:
Determine how to handle null is a responsability of the validation rules. Putting a "continue" on the validator will completely bypass all rules.

bytestream:
@ilvalerione is that more what you have in mind?

ilvalerione:
Rules could include a `nullable` flag to inform the rule if null values must be considered valid or build an error message instead. Something like `allowEmpty` in the `ArrayOf` validation rules: https://github.com/neuron-core/neuron-ai/blob/3.x/src/StructuredOutput/Validation/Rules/ArrayOf.php#L51

bytestream:
@ilvalerione please check now.

PR #604, opened 2026-07-13
Add withFilters to Qdrant and Meilisearch

bytestream:
Elasticsearch, OpenSearch, and Pinecone all have a `withFilters` function that allows you to add filters to the `similaritySearch()` call. This PR copies that behaviour to the Qdrant and Meilisearch vector stores.

I can send another PR with the other vectors updated if you want, or update this PR (but it's a larger change log)?


ilvalerione:
It depends by the capabilities of each database. If it becomes a standard ability we can also think to extend the VectorStoreInterface.

PR #606, opened 2026-07-16
Content from tools

tryvin:
# Allow tools to return rich content (ToolOutput)

## Problem

Tools could only return text (or arrays that got JSON-encoded into strings). There was no way for a tool to hand back richer content — images, documents, audio, etc. — even when the underlying provider APIs natively support such payloads in tool results.

## Solution

Introduce a `ToolOutput` DTO plus an opt-in `HasOutput` interface. Tools that only need text keep their existing `__invoke(): string` signature; tools that need rich content explicitly return a `ToolOutput` carrying `ContentBlockInterface[]`.

The DTO is the **only** signal that the result carries blocks. Arrays are always JSON-encoded (backwards compatible), so there is zero ambiguity — a tool must explicitly opt in via `ToolOutput::blocks([...])` to get block emission.

### `NeuronAI\Tools\ToolOutput` (new)

DTO carrying a tool's structured result.

| Method                                                     | Returns                                                            |
| ---------------------------------------------------------- | ------------------------------------------------------------------ |
| `ToolOutput::text(string $text)`                           | Static factory, text-only result                                   |
| `ToolOutput::blocks(ContentBlockInterface[] $blocks)`      | Static factory, blocks-only result                                 |
| `new ToolOutput(?string $text = null, array $blocks = [])` | Mixed (text + blocks)                                              |
| `hasBlocks()`                                              | `true` when blocks were provided                                   |
| `getBlocks()`                                              | `ContentBlockInterface[]`                                          |
| `getText()`                                                | Explicit text, or concatenation of `TextContent` blocks, or `null` |
| `jsonSerialize()`                                          | `{ text, blocks: [...] }`                                          |

### `NeuronAI\Tools\HasOutput` (new)

Opt-in interface declaring `getOutput(): ToolOutput`. Parallels existing `HasRunKey` / `HasInterrupt` opt-in patterns. Any `ToolInterface` implementation can opt into block emission by implementing `HasOutput` — no need to extend the concrete `Tool` class.

### `Tool` class changes

| Method                           | Behavior                                                                                                                 |
| -------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
| `getResult(): string`            | Unchanged contract on `ToolInterface`. Returns explicit text, or concatenation of `TextContent` blocks. BC.              |
| `setResult(mixed $result): self` | Still off-interface (todo: v4). Accepts `ToolOutput` passthrough; arrays JSON-encoded (BC); everything else string-cast. |
| `getOutput(): ToolOutput`        | New. Declared on `HasOutput`. Returns the structured payload.                                                            |
| `jsonSerialize()`                | Emits both `result` (string, BC) and `resultOutput` (the `ToolOutput` payload).                                          |

### `setResult(mixed)` normalization matrix

| Input                                                | Stored as                                         |
| ---------------------------------------------------- | ------------------------------------------------- |
| `null`, `string`, scalar, `Stringable`               | `ToolOutput::text((string) $x)`                   |
| `array` (any shape — assoc, list, mixed)             | `ToolOutput::text(json_encode($x))` — strict BC   |
| `ToolOutput`                                         | passthrough                                       |
| anything else (incl. single `ContentBlockInterface`) | `(string)` cast → natural error (no `__toString`) |

## Persistence & hydration

`ToolOutput` round-trips through chat history stores (File, SQL, Eloquent).

- **Serialization** (`ToolResultMessage::jsonSerialize()`): for tools implementing `HasOutput` whose output has blocks, the redundant BC `result` string is stripped — `resultOutput` is the canonical payload. This avoids duplicating text content (the same string would otherwise appear in both `result` and `resultOutput.blocks[0].content`).
- **Hydration** (`AbstractChatHistory::deserializeToolCallResult()`): when the persisted `resultOutput` has blocks, blocks are reconstructed via the existing `deserializeContentBlock()` machinery; otherwise the legacy `result` string is read.
- **Legacy data**: histories written before this change only have the `result` string and continue to hydrate as text-only outputs. No migration required.

## Provider support

Mappers honor blocks where the underlying API natively accepts them, and fall back to text otherwise. Detection is via `instanceof HasOutput` — no coupling to the concrete `Tool` class.

| Provider                    | Behavior with blocks                                                                                                        |
| --------------------------- | --------------------------------------------------------------------------------------------------------------------------- |
| **Anthropic**               | `tool_result.content` as native block array (`text`, `image`, `document`) via existing `mapSingleBlock()`                   |
| **AWS Bedrock**             | `toolResult.content` as native block array (`text`, `image`, `document`, `video`, `audio`) via existing `mapContentBlock()` |
| **Gemini**                  | `functionResponse.response.content.parts` array (`text`, `inline_data`, `file_data`)                                        |
| **OpenAI Chat Completions** | `content` as native block array (`text`, `image_url`). Inherited by Cohere, Deepseek, ZAI.                                  |
| **OpenAI Responses**        | `function_call_output.output` as native block array (`input_text`, `input_image`, `input_file`).                            |
| **Mistral**                 | `content` as native block array (`text`, `image_url`, `document_url`, `input_audio`).                                       |
| Ollama                      | Text-only API — falls back to `ToolOutput::getText()`. Unchanged.                                                           |

## Usage example

```php
use NeuronAI\Chat\Enums\SourceType;
use NeuronAI\Chat\Messages\ContentBlocks\ImageContent;
use NeuronAI\Chat\Messages\ContentBlocks\TextContent;
use NeuronAI\Tools\Tool;
use NeuronAI\Tools\ToolOutput;

class CaptureTool extends Tool
{
    public function __invoke(string $target): ToolOutput
    {
        return ToolOutput::blocks([
            new TextContent('Snapshot of ' . $target),
            new ImageContent($this->snapshot($target), SourceType::BASE64, 'image/png'),
        ]);
    }
}
```

For text-only tools, nothing changes:

```php
class SearchTool extends Tool
{
    public function __invoke(string $query): string
    {
        return 'results for ' . $query;
    }
}
```

## Backwards compatibility

- `getResult(): string` — signature and semantics preserved for any non-block input.
- `setResult(['foo' => 'bar'])` — still produces `'{"foo":"bar"}'`.
- `setResult('plain text')` — still produces `'plain text'`.
- `ToolInterface` — unchanged.
- Existing tool implementations, chat history serializers, and stream adapters — behavior unchanged.
- Stream adapters (Vercel AI, AG-UI) intentionally unchanged in v1; they continue to consume `getResult()` (text).

## Tests

36 new tests, all green. Existing tests verified BC.

- `tests/Tools/ToolOutputTest.php` (20 tests) — DTO behavior, `setResult` normalization matrix (string/array/null/scalar/`ToolOutput`), `getResult`/`getOutput` accessors, `jsonSerialize` shape, callable returning string vs. `ToolOutput`.
- `tests/Providers/ToolOutputMapperTest.php` (13 tests) — 6 mappers × {blocks path, BC text path} + a custom `ToolInterface + HasOutput` implementation honored by all mappers.
- `tests/ChatHistory/ToolOutputHydrationTest.php` (3 tests) — round-trip through FileChatHistory with blocks; round-trip with text-only (BC); legacy persisted payload without `resultOutput` still hydrates.

PHPStan level 5: clean. PHP CS Fixer: clean.

## Out of scope (future work)

- Promote `setResult` and `getOutput` to `ToolInterface` in v4 (breaking).
- Extend stream adapters (Vercel AI, AG-UI) to emit block-shaped output once those protocols define it.
- Add block emission for Ollama if/when its tool result schema grows beyond text.

### DIFF
--- a/src/Chat/History/AbstractChatHistory.php
+++ b/src/Chat/History/AbstractChatHistory.php
@@ -23,6 +23,7 @@ use NeuronAI\Chat\Messages\Usage;
 use NeuronAI\Chat\Messages\UserMessage;
 use NeuronAI\Exceptions\ChatHistoryException;
 use NeuronAI\Tools\Tool;
+use NeuronAI\Tools\ToolOutput;
 
 use function array_map;
 use function count;
@@ -195,10 +196,26 @@ abstract class AbstractChatHistory implements ChatHistoryInterface
      */
     protected function deserializeToolCallResult(array $message): ToolResultMessage
     {
-        $tools = array_map(fn (array $tool) => Tool::make($tool['name'], $tool['description'])
-            ->setInputs($tool['inputs'])
-            ->setCallId($tool['callId'])
-            ->setResult($tool['result']), $message['tools']);
+        $tools = array_map(function (array $tool): Tool {
+            $toolInstance = Tool::make($tool['name'], $tool['description'])
+                ->setInputs($tool['inputs'])
+                ->setCallId($tool['callId']);
+
+            if (isset($tool['resultOutput']) && is_array($tool['resultOutput']) && isset($tool['resultOutput']['blocks']) && $tool['resultOutput']['blocks'] !== []) {
+                $blocks = array_map(
+                    $this->deserializeContentBlock(...),
+                    $tool['resultOutput']['blocks']
+                );
+                $toolInstance->setResult(new ToolOutput(
+                    text: $tool['resultOutput']['text'] ?? null,
+                    blocks: $blocks,
+                ));
+            } else {
+                $toolInstance->setResult($tool['result']);
+            }
+
+            return $toolInstance;
+        }, $message['tools']);
 
         return new ToolResultMessage($tools);
     }
--- a/src/Chat/Messages/ToolResultMessage.php
+++ b/src/Chat/Messages/ToolResultMessage.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace NeuronAI\Chat\Messages;
 
+use NeuronAI\Tools\HasOutput;
 use NeuronAI\Tools\ToolInterface;
 use Stringable;
 
@@ -38,7 +39,13 @@ class ToolResultMessage extends UserMessage implements Stringable
             parent::jsonSerialize(),
             [
                 'type' => 'tool_call_result',
-                'tools' => array_map(fn (ToolInterface $tool): array => $tool->jsonSerialize(), $this->tools),
+                'tools' => array_map(function (ToolInterface $tool): array {
+                    $data = $tool->jsonSerialize();
+                    if ($tool instanceof HasOutput && $tool->getOutput()->hasBlocks()) {
+                        unset($data['result']);
+                    }
+                    return $data;
+                }, $this->tools),
             ]
         );
     }
--- a/src/Providers/AWS/MessageMapper.php
+++ b/src/Providers/AWS/MessageMapper.php
@@ -19,6 +19,7 @@ use NeuronAI\Chat\Messages\ToolResultMessage;
 use NeuronAI\Chat\Messages\UserMessage;
 use NeuronAI\Exceptions\ProviderException;
 use NeuronAI\Providers\MessageMapperInterface;
+use NeuronAI\Tools\HasOutput;
 use stdClass;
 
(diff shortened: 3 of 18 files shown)


tryvin:
@ilvalerione Any news?

ilvalerione:
Supported in the next major version v4 currently in beta

PR #610, opened 2026-07-28
Fix VercelAIAdapter text framing for the UI message stream protocol

BrianPortillo:
## The problem

The Vercel AI SDK UI message stream protocol requires every delta of one text part to share a single `id`, framed by `text-start` … `text-end`: https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol#text-parts

`VercelAIAdapter` currently emits each `text-delta` with a freshly generated `id` and never sends the start/end frames. An AI SDK v5+/v6 `useChat` client receives a sequence of unrelated one-fragment parts it cannot assemble, so the streamed text never renders. (Same category as the AG-UI protocol-compliance fix in #608, for the Vercel adapter.)

## The fix

- open a text part on the first delta (`text-start`) and reuse its id for every `text-delta` of that part
- close the open part (`text-end`) before any non-text part begins (tool call, reasoning), so a tool call mid-reply is bracketed correctly
- close a still-open part in `end()` before `finish`/`[DONE]`
- a new text part after a tool call gets a new id

Tool and reasoning frames are unchanged.

## Tests

`test_transform_text_chunk` updated to the framed sequence, and a new `test_text_parts_are_framed_and_share_one_id` covers the full lifecycle: shared delta ids, tool-call bracketing, a new id after reopening, and `end()` closing an open part. Adapter suite: 10 tests, 47 assertions, green.

We run this fix in production against an `ai@6` React `useChat` frontend.

### DIFF
--- a/src/Chat/Messages/Stream/Adapters/VercelAIAdapter.php
+++ b/src/Chat/Messages/Stream/Adapters/VercelAIAdapter.php
@@ -19,6 +19,8 @@ class VercelAIAdapter extends SSEAdapter
 {
     protected bool $started = false;
 
+    protected ?string $currentTextId = null;
+
     /** @var array<string, string> */
     protected array $toolCallIds = [];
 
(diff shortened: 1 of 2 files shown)


BrianPortillo:
Withdrawn.

PR #611, opened 2026-07-28
Add an opt-out for the history trimmer's strict alternation validation

BrianPortillo:
## The problem

`HistoryTrimmer::validateAlternation()` enforces strict user/assistant alternation — a stricter grammar than some providers' own. Anthropic, for example, accepts consecutive same-role messages and merges them into a single turn.

A streamed turn that dies mid-flight (provider error, process restart, closed tab) persists exactly that shape in a database-backed chat history: the user message saved, the assistant reply never written, then the user sends another message. Because the validator runs on every trim, that one dead turn then throws `ChatHistoryException` on every later message — the session is permanently unusable with no way to self-heal.

## The fix

An opt-out constructor flag, default unchanged:

```php
new HistoryTrimmer(strictAlternation: false)
```

When false, the trimmer skips alternation validation and defers to the provider, which is the authority on what it will accept. Trimming behavior itself is untouched.

## Tests

Two added: strict mode still throws on the dead-stream shape (pins the default), lenient mode passes it through to the provider. Suite green.

We run this in production against Anthropic, where the strict validator was bricking chat sessions after mid-stream deploy restarts.

### DIFF
--- a/src/Chat/History/HistoryTrimmer.php
+++ b/src/Chat/History/HistoryTrimmer.php
@@ -35,8 +35,16 @@ class HistoryTrimmer implements HistoryTrimmerInterface
     protected ?int $cachedCount = null;
     protected ?string $cachedLastHash = null;
 
+    /**
+     * @param bool $strictAlternation Throw when the history does not strictly alternate
+     * user/assistant. Providers such as Anthropic accept consecutive same-role messages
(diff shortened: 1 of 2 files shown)


BrianPortillo:
Withdrawn.

PR #612, opened 2026-07-28
Drop a trailing tool call that never received its result on history load

BrianPortillo:
## The problem

A streamed turn that dies between the tool call and the tool result (provider error, process restart, closed tab) persists a trailing `ToolCallMessage` with no matching `ToolResultMessage`. Providers reject that shape on every subsequent request — e.g. Anthropic returns 400 for a `tool_use` block with no `tool_result` — so a session loaded with that tail can never continue: every later message fails, forever.

## The fix

`AbstractChatHistory::deserializeMessages()` now unwinds the one unanswered trailing round on load, so the conversation resumes from the last complete turn. Interior and complete tool rounds are untouched. Because it lives in the shared deserialization path, every store that loads persisted history (File, Eloquent, SQL) gets the heal.

Nothing is deleted from storage — the dangling message is only excluded from the hydrated history the provider sees.

## Tests

Two added via `FileChatHistory` round-trips: a trailing unanswered call is dropped on reload; a complete call→result→answer round survives intact. ChatHistory suite green (the pre-existing `SQLChatHistoryTest` environment errors are identical on the base commit).

We run this heal in production, where mid-stream deaths were permanently bricking database-backed chat threads.

### DIFF
--- a/src/Chat/History/AbstractChatHistory.php
+++ b/src/Chat/History/AbstractChatHistory.php
@@ -25,6 +25,8 @@ use NeuronAI\Exceptions\ChatHistoryException;
 use NeuronAI\Tools\Tool;
 
 use function array_map;
+use function array_pop;
+use function array_values;
 use function count;
 use function end;
 use function is_array;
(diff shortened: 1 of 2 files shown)


BrianPortillo:
Withdrawn.

PR #618, opened 2026-07-29
Code cleanup: non-optimal if conditions.

Denitz:
`if` conditions can be simplified to provide better performance: call functions as the last checks.


ilvalerione:
don't understand it

Denitz:
The statements inside `if` are evaluated in specific order, typically left to right.
if we have `(a && b)` and result of a is false, the statement b is not executed.

So, it's better to have simple statements at the beginning: vars and language constructs like `empty()`

example:
bad: slow function is called first, language construct next. If path is empty, we will still call slow str_starts_with.
`if (!str_starts_with($arg, '-') && empty($options['path']))`

good: fast language construct first, slow function next. If path is empty, we will not call slow function.
`if (empty($options['path']) && !str_starts_with($arg, '-'))`

PR #622, opened 2026-07-29
Code cleanup: redundant type casting

Denitz:
1. `$headers` argument of `SseHttpTransport::buildHeaderString()` is already of type `array<string, string>`
2. `$line` vars in stream handlers are always of `string` type because they are results from `HandleStream::readLine()` method which does return string:
image


ilvalerione:
This is generated by rector. No worry about it.

PR #628, opened 2026-07-29
Code cleanup: ambiguous i modifier

Denitz:
`i` modifier is ambiguous, if a pattern doesn't contain alphabet symbols.

`\w` is case-insensitive itself.

To get the first word from SQL query, it's better to add `u` modifier to grab the first word even in UTF, 'u' add UTF support for `\w`.
Otherwise, the first word in UTF is not captured.


ilvalerione:
Fixed just removing the "i".

I also added tests for both tool to cover this behaviour.

PR #632, opened 2026-07-31
Add --autoload-file option to neuron cli

bytestream:
Ability to run your own autoload file similar to https://phpstan.org/user-guide/discovering-symbols#custom-autoloader and https://docs.phpunit.de/en/13.2/xml-configuration-file.html#the-bootstrap-attribute

### DIFF
--- a/bin/neuron
+++ b/bin/neuron
@@ -3,33 +3,50 @@
 
 declare(strict_types=1);
 
-// Find autoloader
-$autoloadPaths = [
-    __DIR__ . '/../vendor/autoload.php',
-    __DIR__ . '/../../../autoload.php',
-    __DIR__ . '/../../../../autoload.php',
-];
+$autoloadFile = null;
+$args = [$argv[0]];
+for ($i = 1; $i < $argc; $i++) {
+    if ($argv[$i] === '--autoload-file') {
+        if (!isset($argv[$i + 1])) {
+            fwrite(STDERR, 'The --autoload-file option requires a file path' . PHP_EOL);
+            exit(1);
+        }
 
-$autoloaderFound = false;
-foreach ($autoloadPaths as $autoloadPath) {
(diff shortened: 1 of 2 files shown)


ilvalerione:
Thanks for this @bytestream — the option itself makes a lot of sense for the evaluation runner, and the argument parsing/help text look good. One behavioral concern before merging:

**The custom file currently *replaces* the default autoloader discovery instead of supplementing it.**

When `--autoload-file` is passed, the probing of `vendor/autoload.php` is skipped entirely, and the user's file becomes the only thing loaded. But `bin/neuron` still needs Composer's autoloader to load Neuron's own classes (`NeuronCli`, the evaluation runner, etc.). So a bootstrap file that just sets up env vars or constants — which is what the help text suggests ("Load a custom autoloader file before running the command") — would crash with "Class NeuronAI\Console\NeuronCli not found", unless the user knows to `require 'vendor/autoload.php'` inside their own file.

PHPStan can afford replace-style semantics because it ships as a phar with its own bundled autoloader; Neuron runs from the project's Composer install, so I think PHPUnit's `--bootstrap` behavior is the right model here: load the custom file *in addition to* the default autoloader, before it. Since Composer's autoloader is loaded via `require_once`, there's no double-loading risk if the user's bootstrap also requires it.

Two small changes:

1. Require the custom file first (if given), then run the existing default discovery unconditionally:

```php
if ($autoloadFile !== null) {
    if (!file_exists($autoloadFile)) {
        fwrite(STDERR, "Error: autoload file not found: $autoloadFile\n");
        exit(1);
    }
    require_once $autoloadFile;
}

// existing vendor/autoload.php always executed
```

2. That file_exists() check also fixes a second small issue: right now a typo in the path produces a raw PHP fatal error instead of a clean error message like the rest of the script.


Happy to merge once that's adjusted 🙏

PR #642, opened 2026-08-30
feat: add OrcaRouter provider

lovejones2914-spec:
## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

## Description

Neuron users can now point their agent at [OrcaRouter](https://www.orcarouter.ai) with a single line of code, exactly like the existing Deepseek or Grok providers, instead of treating it as an anonymous custom base URL.

**What does this PR do?**

Adds `NeuronAI\Providers\OrcaRouter\OrcaRouter` as a first-class provider. OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding orcarouter as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

**Why is this change needed?**

The provider mirrors how the existing OpenAI-compatible providers are wired: `OrcaRouter` extends `NeuronAI\Providers\OpenAI\OpenAI` and only overrides the base URI, following the same pattern as `XAI\Grok`. The README provider list and the `src/Providers/AGENTS.md` provider table are updated so the provider is discoverable.

**How does this change should be used and documented?**

```php
use NeuronAI\Providers\OrcaRouter\OrcaRouter;

protected function provider(): AIProviderInterface
{
    return new OrcaRouter(
        key: 'ORCAROUTER_API_KEY',
        model: 'openai/gpt-4o-mini',
    );
}
```

## Related Issues

-

## Testing

- [x] All existing tests pass
- [ ] I have tested this with different PHP versions (if applicable)

Added `tests/Providers/OrcaRouterTest.php` covering the chat request payload and the provider base URI. Ran the full suite (1216 tests, 5043 assertions, OK), `phpstan analyse` (level 5, no errors) and `php-cs-fixer` on the new files. Also verified the new provider against the live OrcaRouter endpoint (HTTP 200) using a real key.

## Breaking Changes

- [x] No breaking changes

## Documentation

- [x] No documentation changes needed
- [x] I have updated relevant documentation

## Checklist

## Additional Notes

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ Supported providers:
 - [Mistral](https://docs.neuron-ai.dev/providers/ai-provider#mistral)
 - [HuggingFace](https://docs.neuron-ai.dev/providers/ai-provider#huggingface)
 - [Deepseek](https://docs.neuron-ai.dev/providers/ai-provider#deepseek)
+- [OrcaRouter](https://www.orcarouter.ai) (OpenAI compatible AI gateway)
 - [Grok](https://docs.neuron-ai.dev/providers/ai-provider#grok-x-ai)
 - [AWS Bedrock Runtime](https://docs.neuron-ai.dev/providers/ai-provider#aws-bedrock-runtime)
 - [Cohere](https://docs.neuron-ai.dev/providers/ai-provider#cohere)
--- a/src/Providers/AGENTS.md
+++ b/src/Providers/AGENTS.md
@@ -24,6 +24,7 @@ interface AIProviderInterface {
 | `Ollama/` | Local models |
 | `Mistral/` | Mistral AI |
 | `Deepseek/` | Deepseek |
+| `OrcaRouter/` | OrcaRouter |
 | `Cohere/` | Cohere |
 | `HuggingFace/` | Hugging Face |
 | `XAI/` | xAI Grok |
--- a/src/Providers/OrcaRouter/OrcaRouter.php
+++ b/src/Providers/OrcaRouter/OrcaRouter.php
@@ -1,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace NeuronAI\Providers\OrcaRouter;
+
+use NeuronAI\Providers\OpenAI\OpenAI;
+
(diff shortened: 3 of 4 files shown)


ilvalerione:
For other Open AI compatible providers developers can already use [OpeAILike](https://docs.neuron-ai.dev/providers/ai-provider#openailike)

lovejones2914-spec:
Fair point — `OpenAILike` already covers any OpenAI-compatible provider through a custom base URI, and OrcaRouter is OpenAI-compatible end to end.

The dedicated class is a convenience in the same vein as the existing `Grok` provider, which is also just an `OpenAI` subclass that overrides `$baseUri` (`src/Providers/XAI/Grok.php`). For OrcaRouter specifically, a first-class entry makes the endpoint and model namespace discoverable (`https://api.orcarouter.ai/v1`, models such as `openai/gpt-4o-mini` or `grok-4`) without users having to look up the base URI themselves.

If you'd rather keep the provider surface minimal and have OrcaRouter users configure it via `OpenAILike`, I can drop the new class and the README/AGENTS.md entries and close this PR. Let me know which you prefer.

ilvalerione:
Grok and Deepseek were included because of their mass popularity. OpenAILike was introduced for all others.

lovejones2914-spec:
Understood — first-class provider classes are reserved for broadly adopted APIs (Grok, Deepseek), and OpenAILike is the intended path for everything else that is OpenAI-compatible. OrcaRouter can be configured through OpenAILike with base URI `https://api.orcarouter.ai/v1` and models such as `openai/gpt-4o-mini`, so a dedicated class isn't justified.

Closing this PR as discussed.

Issue #643, opened 2026-08-31
Bedrock streaming crashes on Claude adaptive-thinking responses: reasoning deltas assume `text` and `signature` arrive together


emaferr90:
`src/Providers/AWS/HandleStream.php` handles reasoning deltas like this:

```php
if (isset($event['contentBlockDelta']['delta']['reasoningContent'])) {
    $reasoningChunk = $event['contentBlockDelta']['delta']['reasoningContent']['text'];
    $this->streamState->updateContentBlock(
        $event['contentBlockDelta']['contentBlockIndex'],
        new ReasoningContent($reasoningChunk, $event['contentBlockDelta']['delta']['reasoningContent']['signature'])
    );
    yield new ReasoningChunk($this->streamState->messageId(), $reasoningChunk);
    continue;
}
```

On a real ConverseStream response from a Claude model with (adaptive) thinking enabled — e.g. `eu.anthropic.claude-opus-5` — the reasoning deltas arrive as **separate events**: first a series of `{"reasoningContent":{"text":"…"}}` chunks, then a final `{"reasoningContent":{"signature":"…"}}` (and redacted thinking arrives as `{"reasoningContent":{"redactedContent":"…"}}`). The first `signature`-only (or `redactedContent`) event throws:

```
ErrorException: Undefined array key "text"  (HandleStream.php:76)
```

so **every streamed turn on a thinking-enabled Claude model crashes**.

Suggested fix (what we run in production as a subclass): treat the three keys independently —

```php
if (isset($event['contentBlockDelta']['delta']['reasoningContent'])) {
    $delta = $event['contentBlockDelta']['delta']['reasoningContent'];
    $index = $event['contentBlockDelta']['contentBlockIndex'];

    if (isset($delta['text'])) {
        $this->streamState->updateContentBlock($index, new ReasoningContent($delta['text']));
        yield new ReasoningChunk($this->streamState->messageId(), $delta['text']);
    }

    if (isset($delta['signature'])) {
        // attach the signature to the block accumulated so far
        $this->streamState->signContentBlock($index, $delta['signature']);
    }

    // $delta['redactedContent'] carries no displayable text: accumulate it
    // on the block (or skip) but never index ['text'].
    continue;
}
```

Happy to send a PR with a regression test replaying the real event sequence (text… text, signature) if useful.

Repro: any `stream()` call through `BedrockRuntime` against a Claude model with `additionalModelRequestFields.thinking` not disabled.

[...]

PR #648, opened 2026-09-06
fix(aws): keep redacted reasoning blocks so the next request isn't rejected

emaferr90:
Follow-up to #643, as discussed there.

## Problem

Bedrock's `ReasoningContentBlock` is a union: a block carries either `reasoningText` (`text` + `signature`) or `redactedContent`, the encrypted blob a Claude model returns when its thinking is redacted. 3.16.8 fixed the `reasoningText` half; the blob is still dropped in `HandleStream`, in `HandleChat::mapResponseContent()` and, consequently, in `MessageMapper`.

Anthropic requires the thinking blocks of the *latest* assistant message to be echoed back unmodified — redacted ones included — or the request is rejected:

> Within the latest assistant message, the sequence of consecutive `thinking` blocks must match what the model generated in the original request: you can't rearrange, edit, or partially drop them. This includes `redacted_thinking` blocks.

([Preserving thinking blocks](https://platform.claude.com/docs/en/build-with-claude/thinking#preserving-thinking-blocks), [the matching 400](https://platform.claude.com/docs/en/build-with-claude/thinking-troubleshooting#error-thinking-blocks-modified), [Bedrock's union](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html))

It bites on a tool-use turn that happened to get a redacted block, where the assistant turn has to be replayed alongside the tool result.

## Solution

`RedactedReasoningContent` keeps the blob on a content block, and `MessageMapper` sends it back as `['reasoningContent' => ['redactedContent' => $blob]]`.

| File | Change |
|---|---|
| `Chat/Messages/ContentBlocks/RedactedReasoningContent.php` | new block, mirrors `TextContent` |
| `Chat/Enums/ContentBlockType.php` | `REDACTED_REASONING` case |
| `Chat/History/AbstractChatHistory.php` | deserialization arm — without it the block doesn't survive the history, which is the turn where it matters |
| `Providers/AWS/HandleStream.php` | third branch next to `text` and `signature` |
| `Providers/AWS/HandleChat.php` | same on the non-streaming response |
| `Providers/AWS/MessageMapper.php` | map it back |

## Design notes

- **The new block extends `ContentBlock`, not `TextContent` or `ReasoningContent`.** The blob is not readable text: staying out of those two hierarchies keeps it out of `Message::getContent()`, `getTextBlocks()` and `getReasoning()`, which all filter on them. The cost is that `TokenCounter` falls to its `default` arm (flat ~100 tokens) instead of counting the blob's characters — which seemed the more honest estimate anyway, but say the word and I'll add an explicit arm.
- **No `ReasoningChunk` is yielded** for a redacted delta: there is nothing displayable in it.
- **Other providers are unchanged.** Their mappers hit `default => null` for the new block, i.e. they keep dropping it exactly as today. The Anthropic provider has the same gap (`redacted_thinking` is dropped in its `HandleChat`/`HandleStream`) — happy to send a separate PR if you want it closed too.
- **One existing test moved an index.** In `BedrockReasoningTest` the `redactedContent` delta sat on `contentBlockIndex` 0, the same block as the reasoning text; it now has its own index, because the delta union is per content block and Converse gives each one its own. Left on the shared index, `updateContentBlock()` would concatenate the blob onto the reasoning text.

## Tests

- `BedrockReasoningTest`: both cases now assert the redacted block survives, and that the mapper emits it back between `reasoningText` and `text`.
- `ContentBlockDeserializationTest`: a round trip through the chat history, which is the path that carries the block to the next request.
- `vendor/bin/phpunit` → 1224 tests green (the 7 `WeaviateTest` errors are the service not running locally, unrelated).
- `composer analyse` → no errors. `php-cs-fixer --dry-run` → clean on the touched files.

### DIFF
--- a/src/Chat/Enums/ContentBlockType.php
+++ b/src/Chat/Enums/ContentBlockType.php
@@ -8,6 +8,7 @@ enum ContentBlockType: string
 {
     case TEXT = 'text';
     case REASONING = 'reasoning';
+    case REDACTED_REASONING = 'redacted_reasoning';
     case IMAGE = 'image';
     case FILE = 'file';
     case AUDIO = 'audio';
--- a/src/Chat/History/AbstractChatHistory.php
+++ b/src/Chat/History/AbstractChatHistory.php
@@ -14,6 +14,7 @@ use NeuronAI\Chat\Messages\ContentBlocks\ContentBlockInterface;
 use NeuronAI\Chat\Messages\ContentBlocks\FileContent;
 use NeuronAI\Chat\Messages\ContentBlocks\ImageContent;
 use NeuronAI\Chat\Messages\ContentBlocks\ReasoningContent;
+use NeuronAI\Chat\Messages\ContentBlocks\RedactedReasoningContent;
 use NeuronAI\Chat\Messages\ContentBlocks\TextContent;
 use NeuronAI\Chat\Messages\ContentBlocks\VideoContent;
 use NeuronAI\Chat\Messages\Message;
@@ -258,6 +259,9 @@ abstract class AbstractChatHistory implements ChatHistoryInterface
             ContentBlockType::REASONING => new ReasoningContent(
                 content: $block['content'],
                 id: $block['id'] ?? null
+            ),
+            ContentBlockType::REDACTED_REASONING => new RedactedReasoningContent(
+                content: $block['content']
             ),
             ContentBlockType::IMAGE => new ImageContent(
                 content: $block['content'],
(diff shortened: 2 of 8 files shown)


ilvalerione:
I implemented a solution that doesn't require the introduction of a new content block. It's available in the latest [3.16.11](https://github.com/neuron-core/neuron-ai/releases/tag/3.16.11) 

Thank you for your contributions. Let me know if you find any issue on this.

PR #650, opened 2026-09-07
Allow initializing resources in parallel tool children safely

SmoKE585:
## Type of Change

- [x] 🔧 Improvement/Enhancement

### What does this PR do?

This PR adds an optional `beforeChild` callback to `Agent::parallelToolCalls()`. The callback is passed to `ParallelToolNode` and executed once inside every forked tool task, immediately before the tool itself runs.

Child-initialization failures are caught by the existing task-level error handling and propagated through the normal tool error path instead of becoming an invalid tool result. Existing integrations remain unchanged when the callback is omitted.

Forked child processes may inherit process-bound resources that cannot safely be shared, such as database connections. In a Laravel integration this caused multiple processes to use the same inherited MySQL socket, resulting in protocol errors such as:

```text
Wrong COM_STMT_PREPARE response size. Received 1
```

Framework integrations need a generic hook to reset or initialize these resources independently in every child process.

Executing the initializer through `Spatie\Fork\Fork::before()` is unsafe for this use case because that hook runs outside the protected task callback. If initialization throws, the child may return an empty payload. `unserialize()` then produces `false`, causing a misleading `ToolResultChunk` type error and hiding the original exception.

Running the initializer inside the task `try` block preserves the original error and routes it through Neuron's existing tool error handling.

### Usage

The callback is optional and backward compatible:

```php
$agent->parallelToolCalls(
    true,
    beforeChild: static function (): void {
        // Reset or initialize process-bound resources for this child.
    },
);
```

For example, a Laravel integration can purge the inherited database connection so that the child reconnects lazily when required:

```php
$agent->parallelToolCalls(
    true,
    beforeChild: static function (): void {
        DB::purge();
    },
);
```

The callback is invoked only when multiple tools are actually executed in parallel.

## Related Issues

None.

## Testing

- [x] Tested locally on Linux with PHP 8.3 and `pcntl` enabled.
- [x] Added tests for successful per-child initialization.
- [x] Added a regression test for initializer exception propagation.
- [x] Full test suite passes: 1225 tests, 5100 assertions, 71 skipped.
- [x] Parallel tools suite passes: 8 tests, 27 assertions.
- [x] PHPStan passes with no errors.
- [x] PHP CS Fixer dry run is clean.
- [x] Verified in a Laravel integration with two parallel tools and child-local database connection cleanup.
- [x] Manually validated by a human on a production system with multiple agents, different tool sets, and real chat workflows.

## Breaking Changes

None. The new argument is optional and existing calls remain compatible.

## Documentation

No documentation changes are required. The new optional argument is covered by PHPDoc, usage examples in this PR, and automated tests.

## Checklist

- [x] Code follows the project's coding standards.
- [x] Changes were self-reviewed.
- [x] No new warnings or errors are introduced.
- [x] Commit messages are clear and descriptive.
- [x] The PR addresses one specific goal.

## Additional Notes

The implementation, debugging, Linux `pcntl` reproduction, and test runs were performed with assistance from OpenAI Codex using GPT-5.6 Sol.

### DIFF
--- a/src/Agent/Agent.php
+++ b/src/Agent/Agent.php
@@ -21,6 +21,7 @@ use NeuronAI\Workflow\Persistence\PersistenceInterface;
 use NeuronAI\Workflow\Workflow;
 use NeuronAI\Workflow\WorkflowHandlerInterface;
 use NeuronAI\Workflow\WorkflowState;
+use Closure;
 use Throwable;
 
 use function is_array;
@@ -41,6 +42,8 @@ class Agent extends Workflow implements AgentInterface
 
     protected bool $parallelToolCalls = false;
 
+    protected ?Closure $beforeParallelToolChild = null;
+
     public function init(?InterruptRequest $resumeRequest = null): WorkflowHandlerInterface
     {
         $this->resolveState()->resetToolRuns();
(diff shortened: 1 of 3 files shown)


ilvalerione:
I would also add an `afterChild` callback to complete the hooks.

SmoKE585:
Good point, thanks! I've added an optional afterChild callback. It runs in a finally block after each tool execution, so cleanup is performed even when the tool throws. Exceptions from the callback are handled through the existing task error path. I also added tests for successful cleanup, tool failures, and cleanup callback failures. The full suite passes: 1228 tests, 5107 assertions, 71 skipped.

ilvalerione:
Thank you for the contribution.

PR #668, opened 2026-09-18
fix(zai): stream reasoning chunks

simPod:
## Summary

Stream Z.AI reasoning deltas as reasoning chunks.

Preserve reasoning metadata in normal assistant messages and tool-call continuations.


ilvalerione:
Why use metadata and do not accumulate in a content block? like this: https://github.com/neuron-core/neuron-ai/blob/3.x/src/Providers/OpenAI/Responses/HandleStream.php#L110

simPod:
Thanks for raising this. We need both representations here.

Z.AI requires `reasoning_content` to be returned in the next assistant message, including tool-call continuations. Its [Thinking Mode guide](https://docs.z.ai/guides/capabilities/thinking-mode) says to preserve and return the complete historical `reasoning_content`, and its tool-call example sends it as a top-level assistant field alongside `tool_calls`. The [Chat Completion API](https://docs.z.ai/api-reference/llm/chat-completion) also defines `reasoning_content` separately from `content`.

`ReasoningContent` remains the normalized Neuron representation. Metadata retains Z.AI’s provider-specific transport field because the OpenAI-compatible mapper only serializes text, image, and file blocks; it drops `ReasoningContent` blocks. The cited Responses implementation is protocol-specific: it has separate response item IDs, while Z.AI uses Chat Completions deltas.

I also fixed the synchronous path so it always retains the metadata as well as the normalized block, with a regression test in 6f8535a6.

ilvalerione:
Thanks for clarifying. I agree that Z.AI requires reasoning_content as a separate top-level field, but I’m still unclear why that requires storing a second copy in metadata.
The Z.AI mapper already extracted $message->getReasoning()->content into that field for ordinary messages. Could we keep that approach, add it to mapToolCall(), and accumulate reasoning deltas into ReasoningContent in both streaming hooks? The generic OpenAI mapper dropping reasoning blocks from content shouldn’t prevent the Z.AI mapper from mapping them separately.
Keeping two copies also introduces a consistency issue in the current implementation: with your streaming test fixture, metadata contains "Inspect schema. ", while getReasoning()->content contains only "Inspect ", because the tool-call delta doesn’t update the block.
Is there a case where accumulating into the content block and mapping it this way would lose information?

simPod:
You’re right. The Z.AI mapper can serialize `ReasoningContent` separately, so metadata is not needed. My earlier point about the generic OpenAI mapper was not relevant to the provider-specific mapper.

I changed the implementation to use `ReasoningContent` as the single source of truth. Both streaming hooks now accumulate into the same block, and both ordinary and tool-call messages map that block to the top-level `reasoning_content` field.

I also updated the tests to verify that the complete reasoning value is preserved for synchronous and streamed tool-call continuations.

